PR #1, opened 2015-07-03
fix build with octomap

stonneau:
variable FCL_HAVE_OCTOMAP was actually never activated even if octomap was found,
therefore collision methods were not defined.
Upon activating FCL_HAVE_OCTOMAP when required, there was a need to update the API according to modifications induced in hpp_fcl.


florent-lamiraux:
Merged into HEAD of devel branch.

PR #9, opened 2016-02-15
Provide operator == in CollisionResult, CostSource and Contact

fvalenza:
For the operator in Contact, i am not sure about the "o1 == other.o1 && o2 == other.o2 " part but i think that there is no need to compare the pointed value but to ensure that they point to the same object, and then compare the normal and position of the contact.


jcarpent:
As there is no raw operator == in CollisionGeometry, I think it is the best thing you can do.

PR #11, opened 2016-05-10
modify implementation of operator == in /include/hpp/fcl/math/vec_3fx.h

anna-seppala:
The "equal" function does not return true even in test cases that use the exact same vector as reference and "other". Therefore, I found it reasonable to modify the operator ==.


florent-lamiraux:
Seems good to me.

jmirabel:
Seems good to me too. I hope changing this will not break anything in FCL.

PR #13, opened 2016-06-20
Sync devel branch with master + update test to properly work with boost python

jcarpent:


jmirabel:
I don't really like the `operator==` but these changes correct its expected behaviour.

I think it can be merged.

PR #14, opened 2016-06-28
Added functions to convert an assimp scene to a polyhedron

fvalenza:
Moved assimp module from hpp/pinocchio to hpp-fcl.


jcarpent:
Can you add the lines that makes hpp-fcl still compatible with assimp >= 2.0.0. You can copy past them directly from https://github.com/stack-of-tasks/pinocchio .

include/hpp/fcl/mesh_loader/assimp.h
```diff
@@ -0,0 +1,109 @@
...
+#include <assimp/postprocess.h>
+
+#include <hpp/fcl/BV/OBBRSS.h>
+#include <hpp/fcl/BVH/BVH_model.h>
+
+namespace fcl
+{
+  
+typedef fcl::BVHModel< fcl::OBBRSS > PolyhedronType;
+typedef boost::shared_ptr <PolyhedronType> PolyhedronPtrType;
```
jmirabel: To keep the whole FCL interface homogeneous, I would:
- remove Polyhedron(Ptr)Type typedefs,
- make it a template parameter of the 3 following function.

I mean the following:
```cpp
template <typename BoundingVolume>
void buildMesh(...
  const boost::shared_ptr< BVHModel<BoundingVolume> >&
  )
```

Optionnaly, the shared pointer can be removed as it is not very useful here.

jcarpent:
For support, you need to add the robotpkg ppa to install assimp.

ci/install_linux.sh
```diff
@@ -9,7 +9,7 @@ sudo apt-get -qq --yes --force-yes install libboost-all-dev
 sudo apt-get -qq --yes --force-yes install libccd-dev
 
 # Assimp
-sudo echo "deb [arch=amd64] http://robotpkg.openrobots.org/packages/debian/pub trusty robotpkg" >> /etc/apt/sources.list.d/robotpkg.list
+sudo echo "deb http://robotpkg.openrobots.org/packages/debian/pub precise robotpkg" >> /etc/apt/sources.list.d/robotpkg.list
```
jcarpent: `sudo sh -c "echo \"deb http://robotpkg.openrobots.org/packages/debian/pub precise robotpkg\" >> /etc/apt/sources.list "` may help you.

jcarpent: Or just `sudo apt-get install libassimp-dev`

jmirabel: I agree with @jcarpent , you don't need robotpkg for assimp. Even if the version is not the one you want, it is sufficient for travis.

jcarpent:
Corrected by #15.

PR #18, opened 2016-09-13
Provide operator== for DistanceResult

fvalenza:
See https://github.com/stack-of-tasks/pinocchio/pull/319


jmirabel:
This operator== has been added in Pinocchio by https://github.com/stack-of-tasks/pinocchio/pull/322
It is not mandatory at this stage so I prefer not to add it.

jcarpent:
I just reopened this PR because I think it must be included in the common lib. This features has already been included in the collision result struct and it might be interesting to have for the distance struct also.

PR #25, opened 2017-07-28
Minor fixes

jcarpent:


jmirabel:
`nearest_points` is a vector containing two 3D points.

include/hpp/fcl/collision_data.h
```diff
@@ -469,11 +469,19 @@ struct DistanceResult
...
+                  && b1 == other.b1
+                  && b2 == other.b2;
+
+    if (o1 != NULL xor other.o1 != NULL) return false;
+//    else if (o1 != NULL and other.o1 != NULL) is_same &= *o1 == *other.o1;
+
+    if (o2 != NULL xor other.o2 != NULL) return false;
+//    else if (o2 != NULL and other.o2 != NULL) is_same &= *o2 == *other.o2;
+  
+    return is_same;
```
jmirabel: In the following, the object are not checked any more. Relying on pointer equality is sufficient IMHO so the previous test should be kept.

include/hpp/fcl/collision_data.h
```diff
@@ -469,11 +469,19 @@ struct DistanceResult
...
-    return min_distance == other.min_distance
-        && o1 == other.o1
-        && o2 == other.o2
-        && nearest_points[0] == other.nearest_points[0]
-        && nearest_points[1] == other.nearest_points[1];
+    bool is_same = min_distance == other.min_distance
+                  && nearest_points[0] == other.nearest_points[0]
+                  && nearest_points[1] == other.nearest_points[1]
+                  && b1 == other.b1
+                  && b2 == other.b2;
```
jmirabel: Equality check for b1 and b2 is good.

PR #36, opened 2018-12-27
Add security margin

florent-lamiraux:
This pull request enables users to define a security margin in collision testing.
Two objects are considered in collision if their distance is below the security margin.
The security margin is set in CollisionRequest class.
The modification have been tested with success against all examples in package hpp_benchmark.


florent-lamiraux:
This pull request enables users to define a security margin in collision testing.
Two objects are considered in collision if their distance is below the security margin.
The security margin is set in CollisionRequest class.
The modification have been tested with success against all examples in package hpp_benchmark.

jmirabel:
There are too many changes for me to review. You can merge when it is ready.

PR #50, opened 2019-01-18
Add class for loading meshes from files.

jmirabel:


jmirabel:
The file parsing is handled by assimp, as before. This change adds two classes.
Class `MeshLoader` always calls assimp loading function and create a new GeometryObject.
Class `CachedMeshLoader` keeps a map of (filename, scale) to GeometryObject. If the user request a (filename, scale) that is already in the map, the file is not parsed and the stored GeometryObject is returned.

PR #54, opened 2019-02-06
using flag structure for argurments to collision requests

stonneau:
Added default constructor
Removed default argument to deprecated method to remove ambiguity (behavior remains the same thanks to mentioned default constructor)

### DIFF
--- a/include/hpp/fcl/collision_data.h
+++ b/include/hpp/fcl/collision_data.h
@@ -131,6 +131,14 @@ struct Contact
 
 struct CollisionResult;
 
+/// @brief flag declaration for specifying required params in CollisionResult
+enum CollisionRequestFlag
+{
+  CONTACT               = 0x00001,
+  DISTANCE_LOWER_BOUND  = 0x00002,
+  NO_REQUEST            = 0x01000
+};
+
 /// @brief request to the collision algorithm
 struct CollisionRequest
 {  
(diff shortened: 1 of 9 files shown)


include/hpp/fcl/collision_data.h
```diff
@@ -414,6 +433,28 @@ struct DistanceResult
...
 
+
+inline CollisionRequestFlag operator~(CollisionRequestFlag a)
+{return static_cast<CollisionRequestFlag>(~static_cast<const int>(a));}
+
+inline CollisionRequestFlag operator|(CollisionRequestFlag a, CollisionRequestFlag b)
+{return static_cast<CollisionRequestFlag>(static_cast<const int>(a) | static_cast<const int>(b));}
+
+inline CollisionRequestFlag operator&(CollisionRequestFlag a, CollisionRequestFlag b)
+{return static_cast<CollisionRequestFlag>(static_cast<const int>(a) & static_cast<const int>(b));}
```
jmirabel: I never thought of that trick ! That's amazing.

stonneau: This can probably be templated for all enum types (just googled it:)

emplate<class T> inline T operator~ (T a) { return (T)~(int)a; }
template<class T> inline T operator| (T a, T b) { return (T)((int)a | (int)b); }
template<class T> inline T operator& (T a, T b) { return (T)((int)a & (int)b); }
template<class T> inline T operator^ (T a, T b) { return (T)((int)a ^ (int)b); }
template<class T> inline T& operator|= (T& a, T b) { return (T&)((int&)a |= (int)b); }
template<class T> inline T& operator&= (T& a, T b) { return (T&)((int&)a &= (int)b); }
template<class T> inline T& operator^= (T& a, T b) { return (T&)((int&)a ^= (int)b);

stonneau: http://forums.codeguru.com/showthread.php?553841-DEFINE_ENUM_FLAG_OPERATORS-macro&p=2191205#post2191205

jmirabel: I keep that in mind. Thanks.

PR #61, opened 2019-02-26
obb: fix compatibility with Eigen 3.0.5

jcarpent:


src/BV/OBB.cpp
```diff
@@ -317,7 +317,11 @@ bool obbDisjointAndLowerBoundDistance (const Matrix3f& B, const Vec3f& T,
   // Corner of b axis aligned bounding box the closest to the origin
   Vec3f AABB_corner (T.cwiseAbs () - Bf * b);
   Vec3f diff3 (AABB_corner - a);
+#if EIGEN_VERSION_AT_LEAST(3,1,0)
   diff3 = diff3.cwiseMax (0);
```
jmirabel: Did you check `diff3.array().max(0)` for Eigen < 3.1.0 ? If it works, I would rather avoid the preprocessor if/else.

jcarpent: There is not Scalar overload of `cwiseMax` or `max` with Eigen lower than 3.1.0

jmirabel: The following code compiles fine with Eigen 3.1.0:
```c++
#include <Eigen/Dense>
#include <iostream>

int main()
{
  Eigen::Array3d A(0,1,2);
  std::cout << A.max (1) << '\n';
}
```
so I guess there is an overload in Array class.

jcarpent: But not with 3.0.x

jcarpent: Yes, they have added the overload in 3.1.0

jmirabel: You are right. It does not compile with Eigen 3.0.0.

src/BV/OBB.cpp
```diff
@@ -317,7 +317,8 @@ bool obbDisjointAndLowerBoundDistance (const Matrix3f& B, const Vec3f& T,
   // Corner of b axis aligned bounding box the closest to the origin
   Vec3f AABB_corner (T.cwiseAbs () - Bf * b);
   Vec3f diff3 (AABB_corner - a);
-  diff3 = diff3.cwiseMax (0);
+  diff3 = diff3.cwiseMax (Vec3f::Zero());
```
jmirabel: @jcarpent  Does this suit you ?

I prefer to avoid preprocessor if/else when possible.

jcarpent: Yes, I would have done it at first glance as I was thinking you wanted to keep the initial version.
It is perfect for me now.

jmirabel: Both version boils down to the same assembly code so I give the priority to the most readable solution.

PR #64, opened 2019-03-04
all: add missing #ifdef HPP_FCL_HAVE_OCTOMAP

jcarpent:


jmirabel:
Thanks for the fixes. They make me think that we should just make octomap a required dependency...

PR #66, opened 2019-04-15
Update API of MeshLoader + fix Convex class

jmirabel:
hpp-fcl only supports queries between BVH using the same bounding volume type. The API change of MeshLoader specifies the BV type at construction of the loader rather than when calling the method load.


jmirabel:
I'll handle the Travis failure soon. The failure is actually not a failure, but a degenerate case.

florent-lamiraux:
This is a quite big modification. Does it pass the unit tests ?

jmirabel:
Now yes. You can check Travis status.

PR #67, opened 2019-05-16
Simplify sphereTriangleIntersect

florent-lamiraux:
and add some initializations.

### DIFF
--- a/include/hpp/fcl/collision_data.h
+++ b/include/hpp/fcl/collision_data.h
@@ -348,7 +348,7 @@ public:
                  std::numeric_limits<FCL_REAL>::max()):
   min_distance(min_distance_), o1(NULL), o2(NULL), b1(NONE), b2(NONE)
   {
-    Vec3f nan; nan << sqrt (-1), sqrt (-1), sqrt (-1);
+    Vec3f nan (Vec3f::Constant(std::numeric_limits<FCL_REAL>::quiet_NaN()));
     nearest_points [0] = nearest_points [1] = normal = nan;
   }
 
--- a/include/hpp/fcl/collision_data.h
+++ b/include/hpp/fcl/collision_data.h
@@ -344,12 +344,12 @@ public:
   /// @brief invalid contact primitive information
   static const int NONE = -1;
   
-  DistanceResult(FCL_REAL min_distance_ = std::numeric_limits<FCL_REAL>::max()) : min_distance(min_distance_), 
-                                                                                  o1(NULL),
-                                                                                  o2(NULL),
-                                                                                  b1(NONE),
-                                                                                  b2(NONE)
+  DistanceResult(FCL_REAL min_distance_ =
+                 std::numeric_limits<FCL_REAL>::max()):
+  min_distance(min_distance_), o1(NULL), o2(NULL), b1(NONE), b2(NONE)
   {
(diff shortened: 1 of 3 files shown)


include/hpp/fcl/collision_data.h
```diff
@@ -344,12 +344,12 @@ struct DistanceResult
...
-  DistanceResult(FCL_REAL min_distance_ = std::numeric_limits<FCL_REAL>::max()) : min_distance(min_distance_), 
-                                                                                  o1(NULL),
-                                                                                  o2(NULL),
-                                                                                  b1(NONE),
-                                                                                  b2(NONE)
+  DistanceResult(FCL_REAL min_distance_ =
+                 std::numeric_limits<FCL_REAL>::max()):
+  min_distance(min_distance_), o1(NULL), o2(NULL), b1(NONE), b2(NONE)
   {
+    Vec3f nan; nan << sqrt (-1), sqrt (-1), sqrt (-1);
```
jmirabel: Can you use `Vec3f nan (Vec3f::Constant(std::numeric_limits<FCL_REAL>::quiet_NaN()))` ?

src/narrowphase/details.h
```diff
@@ -349,89 +349,52 @@ namespace fcl {
       const Vec3f& center = tf1.getTranslation();
       const FCL_REAL& radius = s.radius;
       assert (radius >= 0);
-      FCL_REAL eps (std::numeric_limits<FCL_REAL>::epsilon());
-      FCL_REAL radius_with_threshold = radius + eps;
       Vec3f p1_to_center = center - P1;
       FCL_REAL distance_from_plane = p1_to_center.dot(normal);
+      Vec3f closest_point; closest_point << sqrt (-1), sqrt (-1), sqrt (-1);
```
jmirabel: Same here.

src/narrowphase/details.h
```diff
@@ -349,89 +349,52 @@ namespace fcl {
...
-      if(has_contact)
-        {
-          FCL_REAL distance_sqr = center_to_contact.squaredNorm();
+      if (projectInTriangle (P1, P2, P3, normal, center)) {
+        closest_point = center - normal * distance_from_plane;
+        min_distance_sqr = distance_from_plane;
+      } else {
+        // Compute distance to each each and take minimal distance
+        Vec3f nearest_on_edge;
+        min_distance_sqr = segmentSqrDistance(P1, P2, center, closest_point);
```
jmirabel: Cannot `projectInTriangle` know to which edge you are the closest ?

src/narrowphase/details.h
```diff
@@ -2213,14 +2176,15 @@ namespace fcl {
     {
```
jmirabel: I am surprised detecting collision between a plane and a capsule takes so many operations...

src/narrowphase/details.h
florent-lamiraux: Yes. I will do it in a forthcoming commit.

src/narrowphase/details.h
florent-lamiraux: It might be possible to make shorter, but the current implementation seems reasonable to me.

jmirabel:
Looks good to me. You can merge if you are done with testing.

PR #76, opened 2019-09-03
[GJK] Fix GJK::encloseOrigin (fixes unit-tests)

jmirabel:
The failure was due to a NaN value appearing when checking whether the origin is enclosed.

I started [CI on gitlab](https://gepgitlab.laas.fr/humanoid-path-planner/hpp-fcl/pipelines/5447)


jmirabel:
By the way, the first thing EPA does is checking that GJK succeeded in enclosing the origin.
As we always do
```cpp
gjk_res=gjk.evaluate();
if (gjk_res == GJK::Inside) {
  epa.evaluate();
}
```
This check seems useless...

PR #77, opened 2019-09-10
Fix unit-tests and compilation

jmirabel:
Fixes compilation on trusty (see CI of #76 ).
There is a failure on Mac which I do not understand...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,19 +3,37 @@ language: cpp
 env:
   global:
     - CTEST_PARALLEL_LEVEL=4
-  matrix:
-    - BUILD_TYPE=Debug
-    - BUILD_TYPE=Release
+    - CTEST_OUTPUT_ON_FAILURE=1
+    - CXX_FLAGS_DEBUG="-O1"
 
 matrix:
   include:
-    - dist: trusty
-      compiler: gcc
-    - dist: xenial
-      compiler: gcc
-    - dist: xenial
+    - name: "Trusty - Debug - g++"
+      env: BUILD_TYPE=Debug
+      dist: trusty
+      compiler: g++
+
+    - name: "Xenial - Debug - g++"
+      env: BUILD_TYPE=Debug
+      dist: xenial
+      compiler: g++
+    - name: "Xenial - Release - g++"
+      env: BUILD_TYPE=Release
+      dist: xenial
+      compiler: g++
+    - name: "Xenial - Debug - clang"
+      env: BUILD_TYPE=Debug
+      dist: xenial
       compiler: clang
-    - os: osx
+
+    - name: "Bionic - Debug - g++"
+      env: BUILD_TYPE=Debug
+      dist: xenial
+      compiler: g++
+
+    - name: "OSX - Debug - clang"
+      env: BUILD_TYPE=Debug
+      os: osx
       compiler: clang
 
 install:
@@ -29,13 +47,13 @@ script:
   - cd build
 
   # Configure
-  - cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_CXX_FLAGS=-w ..
+  - cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_CXX_FLAGS=-w -DCMAKE_CXX_FLAGS_DEBUG=${CXX_FLAGS_DEBUG} ..
 
   # Build
   - make -j4
 
   # Run unit tests
-  - travis_wait 30 make test || travis_wait 30 ctest -VV
+  - travis_wait 30 ctest
 
   # Make sure we can install and uninstall with no issues
   - sudo make -j4 install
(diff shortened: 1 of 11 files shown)


jmirabel:
@nim65s can you try to understand the linking error on trusty (https://travis-ci.org/humanoid-path-planner/hpp-fcl/jobs/583123627) ?
I am not able to understand it without a computer with trusty.

nim65s:
On 14.04 (libassimp-dev 3.0):
```
╰─>$ objdump -TC /usr/lib/libassimp.so|grep aiScene
00000000001eb280 g    DF .text	00000000000006f0  Base        Assimp::Exporter::ExportToBlob(aiScene const*, char const*, unsigned int)
00000000001eacd0 g    DF .text	00000000000005a2  Base        Assimp::Exporter::Export(aiScene const*, char const*, char const*, unsigned int)
```

And on 16.04 (libassimp-dev 3.2):
```
╰─>$ objdump -TC /usr/lib/x86_64-linux-gnu/libassimp.so|grep aiScene
0000000000151a40 g    DF .text	000000000000063c  Base        aiScene::~aiScene()
00000000001565f0 g    DF .text	0000000000000710  Base        Assimp::Exporter::ExportToBlob(aiScene const*, char const*, unsigned int, Assimp::ExportProperties const*)
0000000000155ef0 g    DF .text	0000000000000700  Base        Assimp::Exporter::Export(aiScene const*, char const*, char const*, unsigned int, Assimp::ExportProperties const*)
0000000000151a40 g    DF .text	000000000000063c  Base        aiScene::~aiScene()
00000000001519b0 g    DF .text	000000000000008e  Base        aiScene::aiScene()
00000000001519b0 g    DF .text	000000000000008e  Base        aiScene::aiScene()
```

So the linker says that it can't find the symbols for `aiScene::~aiScene()`. According to https://github.com/assimp/assimp/commit/5e603d9ed86031c1c554c43ffb102f0df474d823, this was an issue until assimp v3.1

jmirabel:
This is annoying. I don't understand why this was working before...

jcarpent:
> So the linker says that it can't find the symbols for `aiScene::~aiScene()`. According to [assimp/assimp@5e603d9](https://github.com/assimp/assimp/commit/5e603d9ed86031c1c554c43ffb102f0df474d823), this was an issue until assimp v3.1

I think the commit you mentioned is related to Windows.

jcarpent:
@jmirabel Why the lib bacame STATIC?

jmirabel:
> @jmirabel Why the lib bacame STATIC?

Did it ?

In folder tests, the file utility.cpp was compiled by most of the unit-tests so I moved it into a static library called utility.a. I should make compilation a little faster.
But libhpp-fcl shoud should still be a dynamic library...

jmirabel:
> > So the linker says that it can't find the symbols for `aiScene::~aiScene()`. According to [assimp/assimp@5e603d9](https://github.com/assimp/assimp/commit/5e603d9ed86031c1c554c43ffb102f0df474d823), this was an issue until assimp v3.1
> 
> I think the commit you mentioned is related to Windows.

Not necessarily. ASSIMP_API breaks down to `#	define ASSIMP_API __attribute__ ((visibility("default")))`. Without this, on linux, if you pass option `-fvisibility=hidden` to your compiler, then the symbol are *hidden* by default and are public if visibility is set to *default* (*default* means public for the visibility option).

On the version @nim65s mentioned, they pass this option: https://github.com/assimp/assimp/blob/5e603d9ed86031c1c554c43ffb102f0df474d823/CMakeLists.txt#L24

jmirabel:
@nim65s can you try to remove the following line in `src/mesh_loader/assimp.cpp` on 14.04
```
 if (scene) delete scene;
```
?

jmirabel:
The pretty workaround is to include the headers when assimp version is below 3.2. However, I do not understand why this works...

jmirabel:
@jcarpent Does the unit-tests fail on your Mac ?

jmirabel:
For some (yet) unknown reason, `collision --run_test=mesh_mesh` succeeds while `collision` fails on test *mesh_mesh*: https://travis-ci.org/jmirabel/hpp-fcl/builds/584323206

@jcarpent Could you run `collision --log-level=message` to see if you have the same behavior ?

jmirabel:
I understood the issue and I am able to replicate it on my laptop.

jcarpent:
Which is?

jmirabel:
The random number generator is not the same on mac and on linux. All the tests in *collision* generate random values so it is not equivalent to run only mesh_mesh or all the tests in *collision*.

I will dig further to know the exact issue. I suspect that the distance between the 3 pairs of triangles which are problematic is very close to 0. The problem is either in obbDisjoint or in the triangle-triangle distance computation.

jmirabel:
Now it's passing !

jmirabel:
@jcarpent @nim65s thanks for the feedbacks.

PR #78, opened 2019-09-10
Enable Convex / Convex queries + Add Python bindings.

jmirabel:


jmirabel:
contains commit of #77.

jmirabel:
It is not required in FCL but it may be required in downstream packages that stores FCL object with shared pointers. Otherwise, it is not possible to do things like `geomModel.collisionGeometries[0] = fclObject`.

Do you think it is better not to use them ?

jcarpent:
I do not think we should impose the shared_ptr here.
We may have a look to https://www.boost.org/doc/libs/1_57_0/libs/python/doc/v2/register_ptr_to_python.html

jmirabel:
From your link:
>  Note that in order to convert a Python X object to a smart_ptr<X>& (non-const reference), the embedded C++ object must be held by smart_ptr<X>, and that when wrapped objects are created by calling the constructor from Python, how they are held is determined by the HeldType parameter to class_<...> instances. 

I understand that this will work only if the HeldType is a shared pointer. For the case Python case equivalent to `shared_ptr<X> <- X*`, there is no other solution that copying X.

jmirabel:
Rebased onto branch refactoring.

jmirabel:
I remove the changes in EPA for now as they introduced a subtle bug which I have had time to discover.

@jcarpent is your review still valid ?

jmirabel:
@jcarpent I understand you are busy with other stuff.
Could you just let me know whether or not your review is still valid ?

jcarpent:
As I don't have to give it a try, I will merge it. And try to provide fixes if needed. @jmirabel  Is it OK?

jmirabel:
Thanks. We are on branch refactoring, so there is no hurry for checking.

What matters for us is that @LucileRemigy and I are both working on this branch and use this repo to synchronize. The more we wait, the more we diverge.

PR #83, opened 2019-10-09
[CMake] fix octomap detection

nim65s:
Hi,

On some systems, we don't have `OCTOMAP_INCLUDE_DIRS` & `OCTOMAP_LIBRARY_DIRS` but `OCTOMAP_INCLUDEDIR` & `OCTOMAP_LIBDIR`, and therefore we get:

```
-- Checking for module 'octomap >= 1.6'
--   Found octomap , version 1.9.0
-- Pkg-config module octomap v1.9.0 has been detected with success.
-- FCL does not use Octomap
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,7 @@ endif ()
 search_for_boost()
 # Optional dependencies
 add_optional_dependency("octomap >= 1.6")
-if (OCTOMAP_INCLUDE_DIRS AND OCTOMAP_LIBRARY_DIRS)
+if (OCTOMAP_FOUND)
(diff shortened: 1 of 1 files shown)


nim65s:
The next ocurences of `OCTOMAP_INCLUDE_DIRS` and `OCTOMAP_LIBRARY_DIRS` are in this case not working, but it is not an issue in my current use case, as octomap is installed in `/usr` anyway.

In any case, I think we should avoid global `include_directories` and `link_directories` directives, and prefer `PKG_CONFIG_USE_DEPENDENCY` on the targets where this is needed. Maybe this calls for a separate PR.

PR #87, opened 2019-10-22
[bindings] [CMakeLists] Use .so for Mac and .pyd for Windows, fix #86

### LINKED ISSUE #86: [refactoring] Bindings do not work on Mac
The `hppfcl` module does not show up on my Mac after installation.
I use a fairly old OS (10.11, El Capitan), but it will likely be the same on newer versions, as far as I understand.
Apparently, there is a problem with the `.dylib` extension. It should be `.so`, even for Mac.

I followed this post:
https://stackoverflow.com/questions/2488016/how-to-make-python-load-dylib-on-osx

For me, it was enough to do `mv hppfcl.dylib hppfcl.so` inside `lib/python2.7/site-packages` and the problem was solved. This should be handled in the CMakeLists.txt

gabrielebndn:
This closes #86


jmirabel:
Thanks. I updated to the new project policy of the cmake module and used PYTHON_EXT_SUFFIX.

gabrielebndn:
WTH?! You can force-push to MY repo? This is insane! Totally fine with your change, but I need to make a full check of the permissions!

gabrielebndn:
I am sorry, but after the changes it does not work anymore. I still get a dylib

PR #88, opened 2019-10-23
[bindings] [CMakeLists] Fix Mac and Windows library

### LINKED ISSUE #86: [refactoring] Bindings do not work on Mac

gabrielebndn:
Follow-up of #87. Closes #86


jmirabel:
Thank you for this and sorry for having been too quick in merging.

gabrielebndn:
> Thank you for this and sorry for having been too quick in merging.

No grudges :)

PR #89, opened 2019-10-23
[travis] update CI.

jmirabel:


jmirabel:
On all architecture, we use system dependencies only:
- trusty: drop support for Octomap,
- xenial: run travis with Octomap 1.6,
- bionic: run travis with Octomap 1.8.

This reduces a lot CI time since there is no need to compile Octomap anymore.

jmirabel:
CI fails for the following reason:
> Running 1 test case...
> 
> unknown location(0): fatal error in "front_list": memory access violation at address: 0x76be4d020: no mapping at fault address
> 
> /home/travis/build/humanoid-path-planner/hpp-fcl/test/frontlist.cpp(314): last checkpoint
> 
> *** 1 failure detected in test suite "FCL_FRONT_LIST"

or *maybe* more informative

> frontlist: /home/travis/build/humanoid-path-planner/hpp-fcl/include/hpp/fcl/BVH/BVH_model.h:282: const BVNode<BV> &hpp::fcl::BVHModel<hpp::fcl::RSS>::getBV(int) const [BV = hpp::fcl::RSS]: Assertion `id < num_bvs' failed.
> unknown location(0): fatal error in "front_list": signal: SIGABRT (application abort requested)

I skip this error as it is unrelated to this PR. This unit test will be fixed before merging into devel.

PR #90, opened 2019-11-01
[refactoring] Restructure Python tests

gabrielebndn:
The Python tests in [test](https://github.com/humanoid-path-planner/hpp-fcl/tree/8b6d1398376161cd814764ca489357f66000a852/test) were not suitable as unit tests, since they relied on visual inspection of either gepetto-gui or matplotlib plots.
To tidy up, I moved them to a new subfolder [test/scripts](https://github.com/gabrielebndn/hpp-fcl/tree/cb8b64791b70b57c4b49c3ff7f6efabded973091/test/scripts).
Then, I created another subfolder [test/python_unit](https://github.com/gabrielebndn/hpp-fcl/tree/cb8b64791b70b57c4b49c3ff7f6efabded973091/test/python_unit) to hold proper Python unit tests.
In this folder, I added a new unit test [geometric_shapes.py](https://github.com/gabrielebndn/hpp-fcl/blob/cb8b64791b70b57c4b49c3ff7f6efabded973091/test/python_unit/geometric_shapes.py) to verify that all important properties of the different geometric shapes are correctly exposed.
Python unit tests can be listed in [test/python_unit/CMakeLists.txt](https://github.com/gabrielebndn/hpp-fcl/blob/refactoring/test/python_unit/CMakeLists.txt) so that they are automatically executed by the CI when doing `make test`.
In passing, I noticed that the bindings for `Box.halfSide` were not working. This is because objects of type `Vec3f` are not correctly exposed, as they are *not* the same as `Eigen::Vector3d`. I provided a quick fix, but I am not fully satisfied of it.


gabrielebndn:
Please notice that the CI failure was already present and is unrelated to this PR.
Also, I see the Python unit tests are not executed by the CI.
I see `BUILD_PYTHON_INTERFACE` is `OFF` by default. Should we set it to `ON`?

jcarpent:
Yes, sure!

gabrielebndn:
Now the CI fails building because it looks for eigenpy, which is not there... I think it needs to be handled at the robotpkg level. @nim65s, can you do it?

jcarpent:
You should look at Pinocchio travis setup. just copy past and it will work!

gabrielebndn:
> You should look at Pinocchio travis setup. just copy past and it will work!

Ok, I am still confused as to how the CI is handled. Thanks, I'll try!

.travis.yml
```diff
@@ -88,15 +88,8 @@ matrix:
...
-  - sudo sh -c "echo \"deb [arch=amd64] http://robotpkg.openrobots.org/packages/debian/pub $(lsb_release -cs) robotpkg\" >> /etc/apt/sources.list"
-  - curl http://robotpkg.openrobots.org/packages/debian/robotpkg.key | sudo apt-key add -
-  - sudo apt-get update
-  # install eigenpy
-  - sudo apt-get -qqy install robotpkg-py27-eigenpy
-  # set environment variables
-  - export PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:/opt/openrobots/lib/pkgconfig"
-  - export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/opt/openrobots/lib"
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then 'travis_custom/custom_before_install_linux.sh' ; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then 'travis_custom/custom_before_install_linux.sh' ; fi
```
jcarpent: This should be osx

gabrielebndn: Oh, I had overlooked this, thanks!

gabrielebndn:
It should be working now, in the last build the Mac OS test stalled, so I triggered a new run

gabrielebndn:
Now, `BUILD_PYTHON_INTERFACE` is `ON` by default. Consequently, all tests made by the CI include building the Python bindings and install eigenpy.
The CI is now correctly in place and the failure is caused by one unit test, frontlist, which is not related to this PR and does not pass on most linux distributions (apparently it works on trusty and Mac OS).
I think this can be safely merged.

CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ IF(APPLE)
   endif("${isSystemDir}" STREQUAL "-1")
 ENDIF(APPLE)
 
-OPTION(BUILD_PYTHON_INTERFACE OFF)
+OPTION(BUILD_PYTHON_INTERFACE "Build the python bindings" ON)
```
jmirabel: >  I see BUILD_PYTHON_INTERFACE is OFF by default. Should we set it to ON?

CI should set it to ON, regardless of our choice for the default value.
There are installation procedure which relies on hpp-fcl being independent of eigenpy. I would keep the default to OFF and give some time to update this procedure.

python/collision-geometries.cc
```diff
@@ -114,7 +114,9 @@ void exposeShapes ()
     ("Box", init<>())
     .def (init<FCL_REAL,FCL_REAL,FCL_REAL>())
     .def (init<Vec3f>())
-    .def_readwrite ("halfSide", &Box::halfSide)
+    .add_property("halfSide",
+                  make_getter(&Box::halfSide, return_value_policy<return_by_value>()),
+                  make_setter(&Box::halfSide, return_value_policy<return_by_value>()));
```
jmirabel: @gabrielebndn 's comment:
> objects of type Vec3f are not correctly exposed, as they are not the same as Eigen::Vector3d

How are they different ? Vec3f is defined by
`typedef Eigen::Matrix<FCL_REAL, 3, 1> Vec3f;`

gabrielebndn: They are not different, I was wrong when I said that. The point is that it seems `return_internal_reference`, and consequently `def_readwrite`, does not work with eigenpy-bound Eigen objects. In any case, without my modifications, the unit test I am providing was not working, I got a `TypeError`. I'll open an issue on eigenpy to discuss this, but I do not think it is solved easily

gabrielebndn: This issue had already been noticed. See https://github.com/stack-of-tasks/eigenpy/issues/65 and https://github.com/stack-of-tasks/eigenpy/issues/88

CMakeLists.txt
gabrielebndn: When I checked the CI for this PR, the Python unit test was not executed.
If you want to keep it to `OFF` by default for the time being, it's fine for me, but I think there should be at least one CI checking this, and I think this flag should be set to `ON` by the time this branch is merged to devel.
I think I did most of the work required to properly setup the CI for Python test, so any change will be easy to make

jmirabel: 1. set the default to OFF,
2. change CI so that it explicitly use ON for all builds.

.travis.yml
```diff
@@ -5,6 +5,7 @@ env:
     - CTEST_PARALLEL_LEVEL=4
     - CTEST_OUTPUT_ON_FAILURE=1
     - CXX_FLAGS_DEBUG="-O1"
+    - BUILD_PYTHON_INTERFACE=ON
```
jmirabel: I don't think this is sufficient. You need a ` -DBUILD_PYTHON_INTERFACE=${BUILD_PYTHON_INTERFACE}` below.

gabrielebndn: Thanks for the feedback. I think you are right, but let's see what happens. If it does not work, Iĺl try as you say

gabrielebndn: You are right indeed. I'll fix this

gabrielebndn: I have changed it and it seems to work. Let's wait for the CI to finish

CMakeLists.txt
gabrielebndn: Done

jmirabel:
CI failure is not related to this change.
Thanks for the job.

PR #91, opened 2019-11-04
[Travis] Update ci on mac.

jmirabel:
Update Travis on Mac so that it completes faster:
- avoid to install newer versions of git, cmake and boost.
- do not install octomap,
- change compilation flag for faster execution of unit tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,6 @@ matrix:
       env:
         - BUILD_TYPE=Debug
         - CXX_FLAGS_DEBUG="-O3"
-        - BUILD_PYTHON_INTERFACE=OFF
       os: osx
       compiler: clang
       addons:
@@ -87,6 +86,7 @@ matrix:
 
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source travis_custom/custom_before_install_linux.sh ; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then source travis_custom/custom_before_install_osx.sh ; fi
 
 script:
   # Create build directory
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,25 +72,21 @@ matrix:
             - liboctomap-dev
 
     - name: "OSX - Debug - clang"
-      env: BUILD_TYPE=Debug
+      env:
+        - BUILD_TYPE=Debug
+        - CXX_FLAGS_DEBUG="-O3"
+        - BUILD_PYTHON_INTERFACE=OFF
       os: osx
       compiler: clang
       addons:
         homebrew:
-          taps: homebrew/science
           update: true
           packages:
-            - git
-            - cmake
-            - boost
-            - libccd
             - assimp
             - eigen
-            - octomap
 
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source travis_custom/custom_before_install_linux.sh ; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then source travis_custom/custom_before_install_osx.sh ; fi
 
 script:
   # Create build directory
--- a/travis_custom/custom_before_install_osx.sh
+++ b/travis_custom/custom_before_install_osx.sh
@@ -1,3 +1,6 @@
+# Disable homebrew auto-update
+export HOMEBREW_NO_AUTO_UPDATE=1
+
 # Add gepetto tap
 brew tap gepetto/homebrew-gepetto
 


.travis.yml
```diff
@@ -72,25 +72,21 @@ matrix:
...
-            - cmake
-            - boost
-            - libccd
             - assimp
             - eigen
-            - octomap
 
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then source travis_custom/custom_before_install_linux.sh ; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then source travis_custom/custom_before_install_osx.sh ; fi
```
gabrielebndn: You have completely removed this line? Why? I don't think it will work. This file was needed to install eigenpy. The point is brew auto-updates, I think, no idea how to avoid this

gabrielebndn: Oh, so you are not testing the Python bindings on Mac

gabrielebndn: Are you sure there is no other way?

jmirabel: > Are you sure there is no other way?

Not at all. I am a newbie on mac.

gabrielebndn: You can try `HOMEBREW_NO_AUTO_UPDATE=1` : https://github.com/Homebrew/brew/issues/1670

gabrielebndn: It seems I do not have write access to hpp-fcl, so I could not push on your branch. Therefore I made a PR to your repo

.travis.yml
```diff
@@ -72,21 +72,17 @@ matrix:
...
-          taps: homebrew/science
           update: true
           packages:
-            - git
-            - cmake
-            - boost
-            - libccd
             - assimp
             - eigen
-            - octomap
```
gabrielebndn: Why is octomap removed? Because of this, `octomap >= 1.6` is not found by cmake and the corresponding unit test, octree, is not performed by the CI

jmirabel: Because it takes some time to compile it. The unit-test is not run on mac but is run on 16.04 and 18.04.

As of now, I am not really satisfied by the two runs which take more than 10 minutes.

nim65s: FTR, octomap 1.9 is available in robotpkg

gabrielebndn: @nim65s, this is the CI for Mac. We need to use brew, which compiles everything each time.
On Linux distributions, we are using liboctomap-dev. Any reason why we should be using robotpkg instead?

gabrielebndn: > As of now, I am not really satisfied by the two runs on Travis which take more than 10 minutes.

I understand.
- For Mac it seems including eigenpy still takes some time, even if the auto-update is disabled, because brew still installs lots of dependencies, some of which are surprising: cmake, doxygen, icu4c, boost, boost-python, isl, mpfr, libmpc, gcc, openblas, gdbm, openssl@1.1, readline, sqlite, python@2 and numpy@1.16... I would expect at least some of these to be already in the system, no need to reinstall them! But I have no idea how to fix this. So, if it *really* annoys you, maybe you could remove once again the line invoking custom_before_install on Mac, and we won't have Python unit tests for it, until we find a way to speed them up
- For Xenial - debug - clang, the bottleneck is the execution time of the tests. Maybe you could force -03 the same way you did on Mac

nim65s: Here, not, but some other packages require octomap >= 1.9

jcarpent: @gabrielebndn The dependencies are mandatory. You start with a fresh OS and there is no reason to have these dependencies installed. Similar to Windows systems in some sens.

jcarpent: Normally, brew should rely on bottles (binaries packages), so without any compilations (except if the package does not provide the binary). I would advise running the complete pipelines because errors may occur on both sides.

gabrielebndn: @jcarpent I thought that tools like cmake, doxygen, boost, python2 and numpy were already present by default in the freshly-installed system. I took my Mac a lot of time ago so I do not remember all the things I had to install manually and those which were already present. I am surprised they are not.

jmirabel: Does eigen declare a dependency to cmake ?

jcarpent: Yes, but just for the build. In other words, if you use the bottle, it won't install it.

jmirabel: And there is no binary package (bottle) for eigenpy, hence the update of cmake...
Thanks for the explanations !

jmirabel:
I close this. A new PR will be opened when #99 gets merged.

PR #93, opened 2019-11-06
Add version utility methods

gabrielebndn:
1. Added methods `printVersion` and `checkVersionAtLeast`, analogously to Pinocchio
1. Exposed them in Python
1. Added attribute `__version__` to Python module

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,6 +142,7 @@ SET(${PROJECT_NAME}_HEADERS
   include/hpp/fcl/collision_object.h
   include/hpp/fcl/collision_utility.h
   include/hpp/fcl/octree.h
+  include/hpp/fcl/version.h
   include/hpp/fcl/fwd.hh
   include/hpp/fcl/mesh_loader/assimp.h
   include/hpp/fcl/mesh_loader/loader.h
--- a/include/hpp/fcl/version.h
+++ b/include/hpp/fcl/version.h
@@ -1,0 +1,56 @@
+//
+// Copyright (c) 2018 CNRS
+//
+
+#ifndef HPP_FCL_VERSION_H
+#define HPP_FCL_VERSION_H
+
+#include "hpp/fcl/config.hh"
+
+#include <string>
+#include <sstream>
+
+namespace hpp {
+namespace fcl {
+  
+  ///
+  /// \brief Returns the current version of hpp-fcl as a string using
+  ///        the following standard:
+  ///        HPP_FCL_MINOR_VERSION.HPP_FCL_MINOR_VERSION.HPP_FCL_PATCH_VERSION
+  ///
+  inline std::string printVersion(const std::string & delimiter = ".")
+  {
+    std::ostringstream oss;
+    oss
+    << HPP_FCL_MAJOR_VERSION << delimiter
+    << HPP_FCL_MINOR_VERSION << delimiter
+    << HPP_FCL_PATCH_VERSION;
+    return oss.str();
+  }
+  
+  ///
+  /// \brief Checks if the current version of hpp-fcl is at least the version provided
+  ///        by the input arguments.
+  ///
+  /// \param[in] major_version Major version to check.
+  /// \param[in] minor_version Minor version to check.
+  /// \param[in] patch_version Patch version to check.
+  ///
+  /// \returns true if the current version of hpp-fcl is greater than the version provided
+  ///        by the input arguments.
+  ///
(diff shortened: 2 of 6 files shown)


nim65s:
This looks like https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/core/utils/version.hpp 

IMO, this should go to the CMake Submodule, and not in each and every package.

nim65s:
Also, this prevent build in cases where the version computation fails, as

```
oss << CROCODDYL_MAJOR_VERSION << delimiter << CROCODDYL_MINOR_VERSION << delimiter << CROCODDYL_PATCH_VERSION;
```
becomes `oss <<  << delimiter <<  << delimiter << ;`, which is not valid.

gabrielebndn:
> This looks like https://github.com/loco-3d/crocoddyl/blob/master/include/crocoddyl/core/utils/version.hpp
> 
> IMO, this should go to the CMake Submodule, and not in each and every package.

I copied it straight from Pinocchio and I guess the guys working on Crocoddyl did the same.
I do agree that this kind of boilerplate code is likely to be desirable in all of our packages and could well be put inside the cmake submodule.

jmirabel:
> Also, this prevent build in cases where the version computation fails, as
> 
> ```
> oss << CROCODDYL_MAJOR_VERSION << delimiter << CROCODDYL_MINOR_VERSION << delimiter << CROCODDYL_PATCH_VERSION;
> ```
> 
> becomes `oss << << delimiter << << delimiter << ;`, which is not valid.

@gabrielebndn It may be good to enclose portion of the code which depends on those values between ifdefs. The else can be :
- a compilation error that explains clearly what's wrong
- a default version string,
- a exception.

jcarpent:
I will correct it in the JRL cmake module for Pinocchio.

jcarpent:
But, I do not think this should appear by default in the jrl-cmake

gabrielebndn:
@nim65s, the main difficulty I see in putting this in the cmake module comes from the namespaces.

`printVersion` from Pinocchio needs to be in the `pinocchio` namespace.
The one from Crocoddyl needs to be in the `crocoddyl` namespace.
Hpp-FCL is more complicated, it needs to be in the nested namespace `hpp:fcl`.
The same can be said about `checkVersionAtLeast`.

Currently, files `config`, `deprecated` and `warning` only define macros and thus need do not be concerned with namespaces. If we introduce functions, all with the same names, we need them in order to avoid name clashes.

Thus, if we want to introduce support for this in the cmake submodule, we should either make it possible to set a custom (possibly nested) namespace in the CMakeLists.txt, or switch to appropriately-named macros, like `@LIBRARY_NAME@_CHECK_VERSION_AT_LEAST` or `@LIBRARY_NAME@_PRINT_VERSION`.

By the way, I see that a string macro called `@LIBRARY_NAME@_VERSION` is already computed and put in the `config` header. The main differences w.r.t. `printFunction` are
1. If the current commit does not correspond to a release tag, the commit number is added to `@LIBRARY_NAME@_VERSION`. If it the current state does not correspond to a commit, `-dirty` is added.
1. The macro does not allow choosing the delimiter

gabrielebndn:
> Also, this prevent build in cases where the version computation fails, as
> 
> ```
> oss << CROCODDYL_MAJOR_VERSION << delimiter << CROCODDYL_MINOR_VERSION << delimiter << CROCODDYL_PATCH_VERSION;
> ```
> 
> becomes `oss << << delimiter << << delimiter << ;`, which is not valid.

Honestly, I do not understand. In which cases does the version computation fail?

nim65s:
> Honestly, I do not understand. In which cases does the version computation fail?

eg. https://gepgitlab.laas.fr/loco-3d/crocoddyl/merge_requests/258

gabrielebndn:
> > Honestly, I do not understand. In which cases does the version computation fail?
> 
> eg. https://gepgitlab.laas.fr/loco-3d/crocoddyl/merge_requests/258

Yes, that is why https://github.com/jrl-umi3218/jrl-cmakemodules/pull/258 was added. I thought this would solve the problem. Anyhow, I think it was fixed for good with https://github.com/jrl-umi3218/jrl-cmakemodules/pull/270, https://github.com/jrl-umi3218/jrl-cmakemodules/pull/271 and https://github.com/jrl-umi3218/jrl-cmakemodules/pull/273

gabrielebndn:
Following recent discussions, I am closing this PR. I'll wait for https://github.com/jrl-umi3218/jrl-cmakemodules/pull/276 to merge in order to make a new one

PR #94, opened 2019-11-08
Add version support

gabrielebndn:
Updates the cmake module so that `HPP_VERSION_AT_LEAST` and `HPP_VERSION_AT_MOST` are automatically inserted in the `config` header.

Exposes version-related macros in Python as attributes and wrapper functions

### DIFF
--- a/python/version.cc
+++ b/python/version.cc
@@ -4,6 +4,7 @@
 
 #include "hpp/fcl/config.hh"
 #include <boost/python.hpp>
+#include <boost/preprocessor/stringize.hpp>
 
 namespace bp = boost::python;
 
@@ -24,7 +25,9 @@ inline bool checkVersionAtMost(unsigned int major,
 void exposeVersion()
 {
   // Define release numbers of the current hpp-fcl version.
-  bp::scope().attr("__version__") = HPP_FCL_VERSION;
+  bp::scope().attr("__version__") = BOOST_PP_STRINGIZE(HPP_FCL_MAJOR_VERSION) "."
+    BOOST_PP_STRINGIZE(HPP_FCL_MINOR_VERSION) "."
+    BOOST_PP_STRINGIZE(HPP_FCL_PATCH_VERSION);
   bp::scope().attr("HPP_FCL_MAJOR_VERSION") = HPP_FCL_MAJOR_VERSION;
   bp::scope().attr("HPP_FCL_MINOR_VERSION") = HPP_FCL_MINOR_VERSION;
   bp::scope().attr("HPP_FCL_PATCH_VERSION") = HPP_FCL_PATCH_VERSION;
--- a/python/version.cc
+++ b/python/version.cc
@@ -1,0 +1,41 @@
+//
+// Copyright (c) 2019 CNRS
+//
+
+#include "hpp/fcl/config.hh"
+#include <boost/python.hpp>
+
+namespace bp = boost::python;
+
+inline bool checkVersionAtLeast(unsigned int major,
+                                unsigned int minor,
+                                unsigned int patch)
+{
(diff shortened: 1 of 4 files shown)


python/version.cc
```diff
@@ -0,0 +1,41 @@
+//
+// Copyright (c) 2018 CNRS, INRIA
```
nim65s: I am not sure about how this works, but I think you should put only your employer. If someone from INRIA make a significant change to this file, this person should add INRIA too in the author organizations list.

nim65s:
I know your are following what Pinocchio is doing, but I still see this C++ style for python function definitions as a mistake. Python users will probably look for `check_version_at_most` instead of `checkVersionAtMost`, and if you actually use this version in a python files, linter will complain about the bad style.

nim65s:
ref: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

python/version.cc
gabrielebndn: I admit I just blindly copy-pasted it. Thanks for the remark, I'll fix it

jcarpent: But if it is really copy past, then you should keep the original header license. This is the rule.

jcarpent:
All the naming are coming from C++, so we should stick on it, because we are making bindings.

gabrielebndn:
I do not think we should be blinded by the usual Python conventions. From PEP8 itself: https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

Quoting:
> A style guide is about consistency. Consistency with this style guide is important. Consistency within a project is more important. Consistency within one module or function is the most important.

> do not break backwards compatibility just to comply with this PEP!

I our case, it makes much more sense to keep using C++ style function naming, IMO

python/version.cc
gabrielebndn: The file is now different to the point it has little to do with the original.

jcarpent: As the original author of the copy-past file, I'm may feel a bit sad to my the authorship removed. I mean here is just a matter of politeness/kindness I would say, before just rules.

python/version.cc
```diff
@@ -0,0 +1,41 @@
...
+                               unsigned int minor,
+                               unsigned int patch)
+{
+  return HPP_FCL_VERSION_AT_MOST(major, minor, patch);
+}
+
+void exposeVersion()
+{
+  // Define release numbers of the current hpp-fcl version.
+  bp::scope().attr("__version__") = HPP_FCL_VERSION;
```
jcarpent: Printing the version with dirty elements inside may not be a good idea.

gabrielebndn: Yeah... I thought about it... But then what is the C++ macro for? Is it a good idea to include the dirty elements in the C++ string?

jcarpent: You have to put "X.Y.Z"

gabrielebndn: I confirm that, if the string `__version__` contains commit information, it does not comply to [PEP 440](https://www.python.org/dev/peps/pep-0440/) and this creates problems for packages and methods expecting strict compliance such as [packaging.version.parse](https://packaging.pypa.io/en/latest/version/#packaging.version.parse), as it is mistaken for a legacy version
1. Should I change this
1. Should we modify the C++ macro so that it does not add the extra info? After all, the resulting version string is highly unconventional. Maybe something like `PKG_VERSION` and `PACKAGE_VERSION_FULL`?

jmirabel: I suggest to follow PEP 440. No need to change the C++ macro.

gabrielebndn:
Thanks @jmirabel for your commit

gabrielebndn:
I think everything is fine, here. Can we merge this?

PR #95, opened 2019-11-10
assimp: fix issue with recent version of assimp

jcarpent:
There is a compilation due to recent versions of assimp. This PR fixes it temporarly.
The current PR on the offical assimp repository with the fix is here: https://github.com/assimp/assimp/pull/2758


jmirabel:
Thank you Justin.

PR #96, opened 2019-11-10
Travis: change policy to only perform build in DEBUG mode on Bionic

jcarpent:


jcarpent:
needs #95 to work on OSX.

.travis.yml
```diff
@@ -32,11 +36,11 @@ script:
   - cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_CXX_FLAGS=-w ..
 
   # Build
-  - make -j4
+  - make -j2
 
   # Run unit tests
-  - travis_wait 30 make test || travis_wait 30 ctest -VV
+  - travis_wait 30 make -j2 test
```
jmirabel: This overrides `CTEST_PARALLEL_LEVEL=4` right ?
If we delete `ctest -VV`, we should add `--output-on-failure` so that we have some information when it fails.

jcarpent: Not really, I think we should use the variable `MAKEFLAGS` for that. The `-j2` is used for compiling the test.
For the second point, we can use `CTEST_OUTPUT_ON_FAILURE=1

jcarpent:
I've just rebase from the devel branch.

.travis.yml
jmirabel: > Not really, I think we should use the variable MAKEFLAGS for that. The -j2 is used for compiling the test.

You are right.

PR #97, opened 2019-11-12
New folder Internal for internal header

LucileRemigy:
new folder "internal" in hpp/fcl for internal header + traits traversal for collision and distance


src/traits_traversal.h
```diff
@@ -1,96 +1,91 @@
...
- *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
- *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
- *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
- *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
- *  POSSIBILITY OF SUCH DAMAGE.
- */
-
-/** \author Florent Lamiraux */
-
-
```
jmirabel: Would you mind adding the following header to this file:
```
/*
 * Software License Agreement (BSD License)
 *
 *  Copyright (c) 2019, CNRS-LAAS
 *  All rights reserved.
 */

/** \author Lucile Remigy, Joseph Mirabel */
```

src/traits_traversal.h
```diff
@@ -1,96 +1,91 @@
...
+#include <../src/collision_node.h>
+#include <hpp/fcl/internal/traversal_node_setup.h>
+#include <../src/distance_func_matrix.h>
 
 namespace hpp
 {
 namespace fcl
 {
+
+#ifdef HPP_FCL_HAVE_OCTOMAP
```
jmirabel: This pre-processor check if not at the correct place. It should only encapsulate OcTree template specialization.

LucileRemigy: Do I delete the "ifdef hpp_fcl_have_octomap" ?

jmirabel: No. The ifdef 
> should only encapsulate OcTree template specializations.

This means all the code like
```cpp
template <>
TraversalTraitsCollision<...OcTree...>
{
...
};
```

LucileRemigy: is it better with the new push?

jmirabel: Yes, perfect. Thank you.

Issue #92, opened 2019-11-05
Merging the refactoring branch to devel


gabrielebndn:
I am writing this issue to keep track of what is missing for the refactoring branch to be mature enough for a release.
So far, I can identify:
- [x] Handle redundancy between `lz` and `HalfLength` in `Cylinder`, `Capsule` and `Cone`. This can be done by either removing `lz` or by making the variables private and providing appropriate getter/setter methods which ensure consistency between the two values
- [x] Fix the CI failure in unit test `frontlist`. As far as I have understood, the buggy feature could be removed altogether
- [x] Expand the Python bindings so that they cover at least all classes and methods currently exposed in Pinocchio and the most important core algorithms
- [ ] Verify the Python bindings through appropriate unit tests
- [x] Install internal headers.
- [x] Run integration test for known dependencies (pinocchio and HPP)

@jmirabel, I know you have something to add to the list. Feel free to edit this post
@nmansard, could you provide feedback of what you need to be exposed for your tutorials?

jmirabel:
@LucileRemigy could you take care of the first item ?

jmirabel:
The integration test and the Python bindings unit-tests are not blocking the merging of refactoring into devel. I will do the merge before doing the integration test so that people can start playing with the new features and eventually report issues.

jcarpent:
Test are passing well on my machine. You can proceed to the merging I think.

jmirabel:
I tested it on HPP. There is one tricky change in class Transform3f which had impact in 2/3 places. The code compiles and I think the semantic is correct. I didn't run the benchmarks. I merge this now.

gabrielebndn:
I am fixing a couple of things. Please wait a minute

gabrielebndn:
The refactoring branch has been merged. All points are satisfied, except for the unit tests for the Python bindings. There are already tests for the bindings which are interesting to me, the rest can be addressed in the long term. I think this issue can be closed.

PR #98, opened 2019-11-13
Delete lz for cone cylinder and Capsule, now used HalfLength

LucileRemigy:


jmirabel:
This solves one point of #92.

gabrielebndn:
Thanks for this job!
I recommend taking care of the Python bindings too and of the corresponding unit test.
You should enable building them by setting `BUILD_PYTHON_INTERFACE` to `ON`

include/hpp/fcl/shape/geometric_shapes.h
```diff
@@ -151,18 +151,14 @@ class Sphere : public ShapeBase
 class Capsule : public ShapeBase
 {
 public:
-  Capsule(FCL_REAL radius_, FCL_REAL lz_) : ShapeBase(), radius(radius_), lz(lz_)
+  Capsule(FCL_REAL radius_, FCL_REAL lz_) : ShapeBase(), radius(radius_)
   {
-    lz = lz_;
     HalfLength = lz_/2;
```
jmirabel: for homogeneity, could you rename _**H**alfLength_ into _**h**alfLength_ ?

include/hpp/fcl/shape/geometric_shapes.h
```diff
@@ -174,15 +170,15 @@ class Capsule : public ShapeBase
...
   }
 
   Matrix3f computeMomentofInertia() const
   {
-    FCL_REAL v_cyl = radius * radius * lz * boost::math::constants::pi<FCL_REAL>();
+    FCL_REAL v_cyl = radius * radius * (HalfLength * 2) * boost::math::constants::pi<FCL_REAL>();
     FCL_REAL v_sph = radius * radius * radius * boost::math::constants::pi<FCL_REAL>() * 4 / 3.0;
     
-    FCL_REAL ix = v_cyl * lz * lz / 12.0 + 0.25 * v_cyl * radius + 0.4 * v_sph * radius * radius + 0.25 * v_sph * lz * lz;
+    FCL_REAL ix = v_cyl * (HalfLength * 2) * (HalfLength * 2) / 12.0 + 0.25 * v_cyl * radius + 0.4 * v_sph * radius * radius + 0.25 * v_sph * (HalfLength * 2) * (HalfLength * 2);
```
jmirabel: Replace this line by:

    FCL_REAL ix = v_cyl * HalfLength * HalfLength / 3. + 0.25 * v_cyl * radius + 0.4 * v_sph * radius * radius + v_sph * HalfLength * HalfLength;

LucileRemigy:
i'm working on compilation errors

include/hpp/fcl/shape/geometric_shapes.h
```diff
@@ -219,13 +211,13 @@ class Cone : public ShapeBase
...
   {
-    return boost::math::constants::pi<FCL_REAL>() * radius * radius * lz / 3;
+    return boost::math::constants::pi<FCL_REAL>() * radius * radius * (HalfLength * 2) / 3;
   }
 
   Matrix3f computeMomentofInertia() const
   {
     FCL_REAL V = computeVolume();
-    FCL_REAL ix = V * (0.1 * lz * lz + 3 * radius * radius / 20);
+    FCL_REAL ix = V * (0.1 * (HalfLength * 2) * (HalfLength * 2) + 3 * radius * radius / 20);
```
jmirabel: Use this line:

    FCL_REAL ix = V * (4 / 10 * HalfLength * HalfLength + 3 * radius * radius / 20);

and add, as a doxygen comment of computeMomentofInertia

    /// \todo verify this formula as it seems different from https://en.wikipedia.org/wiki/List_of_moments_of_inertia#List_of_3D_inertia_tensors

include/hpp/fcl/shape/geometric_shapes.h
gabrielebndn: This could be:
`v_cyl * HalfLength * HalfLength / 3.0 + 0.25 * v_cyl * radius + 0.4 * v_sph * radius * radius + v_sph * HalfLength * HalfLength;`

include/hpp/fcl/shape/geometric_shapes.h
```diff
@@ -266,13 +254,13 @@ class Cylinder : public ShapeBase
...
   {
-    return boost::math::constants::pi<FCL_REAL>() * radius * radius * lz;
+    return boost::math::constants::pi<FCL_REAL>() * radius * radius * (HalfLength * 2);
   }
 
   Matrix3f computeMomentofInertia() const
   {
     FCL_REAL V = computeVolume();
-    FCL_REAL ix = V * (3 * radius * radius + lz * lz) / 12;
+    FCL_REAL ix = V * (3 * radius * radius + (HalfLength * 2) * (HalfLength * 2)) / 12;
```
jmirabel: FCL_REAL ix = V * (3 * radius * radius + HalfLength * HalfLength / 3;

gabrielebndn:
I see that in lots of places `lz` was simply replaced by `HalfLength * 2`. This introduces an extra multiplication and kinds of defeats the purpose of using `HalfLength` instead of `lz`, I think.
I do not know whether the compiler is smart enough to figure it out, but in many places the `*2` can be conflated to other constants. See [my comment](https://github.com/humanoid-path-planner/hpp-fcl/pull/98#discussion_r345757209) for an example

src/narrowphase/details.h
```diff
@@ -73,8 +73,8 @@ namespace fcl {
        const Capsule& s2, const Transform3f& tf2,
        Vec3f* contact_points, FCL_REAL* penetration_depth, Vec3f* normal_)
     {
-      Vec3f pos1 (tf2.transform (Vec3f (0., 0., 0.5 * s2.lz))); // from distance function
-      Vec3f pos2 (tf2.transform (Vec3f (0., 0., -0.5 * s2.lz)));
+      Vec3f pos1 (tf2.transform (Vec3f (0., 0., 0.5 * (s2.HalfLength * 2)))); // from distance function
+      Vec3f pos2 (tf2.transform (Vec3f (0., 0., -0.5 * (s2.HalfLength * 2))));
```
jmirabel: Vec3f pos1 (tf2.transform (Vec3f (0., 0.,  s2.HalfLength))); // from distance function
    Vec3f pos2 (tf2.transform (Vec3f (0., 0., -s2.HalfLength)));

src/narrowphase/details.h
```diff
@@ -108,8 +108,8 @@ namespace fcl {
        const Capsule& s2, const Transform3f& tf2,
        FCL_REAL& dist, Vec3f& p1, Vec3f& p2, Vec3f& normal)
     {
-      Vec3f pos1 (tf2.transform (Vec3f (0., 0., 0.5 * s2.lz)));
-      Vec3f pos2 (tf2.transform (Vec3f (0., 0., -0.5 * s2.lz)));
+      Vec3f pos1 (tf2.transform (Vec3f (0., 0., 0.5 * (s2.HalfLength * 2))));
+      Vec3f pos2 (tf2.transform (Vec3f (0., 0., -0.5 * (s2.HalfLength * 2))));
```
jmirabel: same as above.

src/shape/geometric_shapes_utility.cpp
```diff
@@ -303,9 +303,9 @@ void computeBV<AABB, Cylinder>(const Cylinder& s, const Transform3f& tf, AABB& b
   const Matrix3f& R = tf.getRotation();
   const Vec3f& T = tf.getTranslation();
 
-  FCL_REAL x_range = fabs(R(0, 0) * s.radius) + fabs(R(0, 1) * s.radius) + 0.5 * fabs(R(0, 2) * s.lz);
-  FCL_REAL y_range = fabs(R(1, 0) * s.radius) + fabs(R(1, 1) * s.radius) + 0.5 * fabs(R(1, 2) * s.lz);
-  FCL_REAL z_range = fabs(R(2, 0) * s.radius) + fabs(R(2, 1) * s.radius) + 0.5 * fabs(R(2, 2) * s.lz);
+  FCL_REAL x_range = fabs(R(0, 0) * s.radius) + fabs(R(0, 1) * s.radius) + 0.5 * fabs(R(0, 2) * (s.HalfLength * 2));
+  FCL_REAL y_range = fabs(R(1, 0) * s.radius) + fabs(R(1, 1) * s.radius) + 0.5 * fabs(R(1, 2) * (s.HalfLength * 2));
+  FCL_REAL z_range = fabs(R(2, 0) * s.radius) + fabs(R(2, 1) * s.radius) + 0.5 * fabs(R(2, 2) * (s.HalfLength * 2));
```
jmirabel: Merge the real factors 0.5 and 2 here.

include/hpp/fcl/shape/geometric_shapes.h
```diff
@@ -235,28 +228,24 @@ class Cone : public ShapeBase
 
   Vec3f computeCOM() const
   {
-    return Vec3f(0, 0, -0.25 * lz);
+    return Vec3f(0, 0, -0.25 * (halfLength * 2));
```
jmirabel: 0.25 * 2 -> 0.5

src/narrowphase/details.h
```diff
@@ -2346,8 +2346,8 @@ namespace fcl {
           else
             {
               if (d < 0) normal = new_s2.n; else normal = -new_s2.n;
-              p1 = p2 = T - dir_z * (0.5 * s1.lz) +
-                dir_z * (-0.5 * distance / s1.radius * s1.lz) - new_s2.n * d;
+              p1 = p2 = T - dir_z * (s1.halfLength) +
+                dir_z * (-0.5 * distance / s1.radius * (s1.halfLength * 2)) - new_s2.n * d;
```
jmirabel: dir_z * (- distance / s1.radius * s1.halfLength) - new_s2.n * d;

src/shape/geometric_shapes_utility.cpp
```diff
@@ -277,7 +277,7 @@ void computeBV<AABB, Capsule>(const Capsule& s, const Transform3f& tf, AABB& bv)
   const Matrix3f& R = tf.getRotation();
   const Vec3f& T = tf.getTranslation();
 
-  Vec3f v_delta(0.5 * (R.col(2)*s.lz).cwiseAbs() + Vec3f::Constant(s.radius));
+  Vec3f v_delta(0.5 * (R.col(2)* (s.halfLength * 2)).cwiseAbs() + Vec3f::Constant(s.radius));
```
jmirabel: Vec3f v_delta(R.col(2).cwiseAbs() * s.halfLength + Vec3f::Constant(s.radius));

src/shape/geometric_shapes_utility.cpp
```diff
@@ -288,9 +288,9 @@ void computeBV<AABB, Cone>(const Cone& s, const Transform3f& tf, AABB& bv)
   const Matrix3f& R = tf.getRotation();
   const Vec3f& T = tf.getTranslation();
 
-  FCL_REAL x_range = fabs(R(0, 0) * s.radius) + fabs(R(0, 1) * s.radius) + 0.5 * fabs(R(0, 2) * s.lz);
-  FCL_REAL y_range = fabs(R(1, 0) * s.radius) + fabs(R(1, 1) * s.radius) + 0.5 * fabs(R(1, 2) * s.lz);
-  FCL_REAL z_range = fabs(R(2, 0) * s.radius) + fabs(R(2, 1) * s.radius) + 0.5 * fabs(R(2, 2) * s.lz);
+  FCL_REAL x_range = fabs(R(0, 0) * s.radius) + fabs(R(0, 1) * s.radius) + 0.5 * fabs(R(0, 2) * (s.halfLength * 2));
+  FCL_REAL y_range = fabs(R(1, 0) * s.radius) + fabs(R(1, 1) * s.radius) + 0.5 * fabs(R(1, 2) * (s.halfLength * 2));
+  FCL_REAL z_range = fabs(R(2, 0) * s.radius) + fabs(R(2, 1) * s.radius) + 0.5 * fabs(R(2, 2) * (s.halfLength * 2));
```
jmirabel: 0.5 * 2 -> 1.

include/hpp/fcl/shape/geometric_shapes.h
```diff
@@ -266,13 +255,13 @@ class Cylinder : public ShapeBase
...
   {
-    return boost::math::constants::pi<FCL_REAL>() * radius * radius * lz;
+    return boost::math::constants::pi<FCL_REAL>() * radius * radius * (halfLength * 2);
   }
 
   Matrix3f computeMomentofInertia() const
   {
     FCL_REAL V = computeVolume();
-    FCL_REAL ix = V * (3 * radius * radius + lz * lz) / 12;
+    FCL_REAL ix = V * (3 * radius * radius + halfLength * halfLength / 3);
```
gabrielebndn: This is mistake! It should be either
` V * (3 * radius * radius + 4 * halfLength * halfLength) / 12;`
or
` V * (radius * radius / 4 + halfLength * halfLength / 3);`
I cannot say which one is faster, though

include/hpp/fcl/shape/geometric_shapes.h
```diff
@@ -219,13 +211,14 @@ class Cone : public ShapeBase
...
-    return boost::math::constants::pi<FCL_REAL>() * radius * radius * lz / 3;
+    return boost::math::constants::pi<FCL_REAL>() * radius * radius * (halfLength * 2) / 3;
   }
 
+  /// \todo verify this formula as it seems different from https://en.wikipedia.org/wiki/List_of_moments_of_inertia#List_of_3D_inertia_tensors
   Matrix3f computeMomentofInertia() const
   {
     FCL_REAL V = computeVolume();
-    FCL_REAL ix = V * (0.1 * lz * lz + 3 * radius * radius / 20);
+    FCL_REAL ix = V * (4 / 10 * halfLength * halfLength + 3 * radius * radius / 20);
```
gabrielebndn: `4 / 10` is not zero? I would write `0.4`

include/hpp/fcl/shape/geometric_shapes.h
jmirabel: V * (radius * radius / 4 + halfLength * halfLength / 3);

should be faster.

include/hpp/fcl/shape/geometric_shapes.h
jmirabel: Indeed, 0.4 is safer. My bad.

src/narrowphase/details.h
```diff
@@ -143,15 +143,15 @@ namespace fcl {
...
       Vec3f B (tf2.transform (Vec3f (0, 0,  lz2)));
       // Position of the center of the sphere
       Vec3f S (tf1.getTranslation ());
       // axis of the cylinder
       Vec3f u (tf2.getRotation ().col (2));
-      assert ((B - A - s2.lz * u).norm () < eps);
+      assert ((B - A - (s2.halfLength * 2) * u).norm () < eps);
       Vec3f AS (S - A);
       // abscissa of S on cylinder axis with A as the origin
       FCL_REAL s (u.dot (AS));
```
jmirabel: A note for later: a tiny performance improvement could be achieved using the abscissa with S as the origin.

LucileRemigy: I add a comment to the todolist?

jmirabel: Good idea.

gabrielebndn:
Looks fine to me!

include/hpp/fcl/shape/geometric_shapes.h
gabrielebndn: Indeed it does seem different for Wikipedia. This should not affect this PR, though

gabrielebndn: @jmirabel, the formula here expresses the inertia moment around an axis passing through the base. Wikipedia's formula expresses around an axis passing through the tip.

https://math.stackexchange.com/questions/1082670/moment-of-inertia-of-a-solid-cone-about-a-diameter-of-the-base-as-axis

I think neither of them are useful: inertia moments should be expressed around an axis passing through the CoM, in my opinion. This is what Pinocchio does, for instance.
Regardless, you can remove this comment and maybe replace it with some documentation.
I think I will update Wikipedia to clarify

gabrielebndn: I figured it out. The doc is in the superclass `CollisionGeometry`, in [include/hpp/fcl/collision_object.h](https://github.com/humanoid-path-planner/hpp-fcl/blob/6b10c369eabc1c714d87c6409bb79b3824c89553/include/hpp/fcl/collision_object.h).

For all objects, `computeMomentofInertia` computes the inertia tensor w.r.t. the origin of the shape. For a cone, the origin is halfway along the axis; since the CoM is halfway between the origin and the base, the moment of inertia about an axis passing through the origin or through the base have the same value. This makes sense. Then, the inertia tensor w.r.t. to the CoM is automatically computed with `computeMomentofInertiaRelatedToCOM` using the parallel-axis theorem.

jmirabel: See #104

PR #99, opened 2019-11-14
[Python] Update bindings with requirements from Pinocchio.

jmirabel:


jmirabel:
Solves another point of #92

gabrielebndn:
Since I see you have exposed `std::vector<Contact>`, I think we should expose `std::vector` for the other classes too.

jmirabel:
I exposed it because it is used by class CollisionResult. The other vectors are not used. I have nothing against adding them here.
I am just not sure where they fit best: in hpp-fcl or in pinocchio ?

gabrielebndn:
> I exposed it because it is used by class CollisionResult. The other vectors are not used. I have nothing against adding them here.
> I am just not sure where they fit best: in hpp-fcl or in pinocchio ?

I am referring to the ones which are already exposed in Pinocchio. Since they are an hpp-fcl-related thing, and there already is `StdVec_Contact` exposed in hpp-fcl, it seems reasonable to me to move them to hpp-fcl, for consistency. It is purely a design choice.

jmirabel:
Travis still complains about the frontlist unit test.

PR #101, opened 2019-11-14
Bug fix in propagateBVHFrontListCollisionRecurse + unit-tests

jmirabel:
This solves one point of #92


gabrielebndn:
By the way, you had noticed a warning about bad indentation which actually was a bug, right?

jmirabel:
> By the way, you had noticed a warning about bad indentation which actually was a bug, right?

When I look at the code, it isn't a bug. But the compilation warning should be fixed.

jmirabel:
CI is happy now !

PR #102, opened 2019-11-14
[CMake] ccd is required to run tests

nim65s:


jmirabel:
ccd is not a dependency of hpp-fcl. What is the issue exactly ?

nim65s:
I don't have any issue, I'm just trying to check dependencies consistencies.
Here, we have:
- https://github.com/humanoid-path-planner/hpp-fcl/blob/6b10c369eabc1c714d87c6409bb79b3824c89553/INSTALL#L9
- some references in tests, like https://github.com/humanoid-path-planner/hpp-fcl/blob/6b10c369eabc1c714d87c6409bb79b3824c89553/test/test_fcl_shape_mesh_consistency.cpp#L291
- https://github.com/Gepetto/robotpkg/blob/b6c9b3d45870ecbeb6821a28a88da38215e583fa/path/hpp-fcl/Makefile#L25

So I thought this was a dependency.
But if I'm wrong, we can correct all those other places instead.

jmirabel:
The dependency was removed long ago. I cannot find the commit but I would say at least 3 years ago.

There are still some erroneous mention of ccd which ought to be removed...

nim65s:
Ok, thanks :)

PR #104, opened 2019-11-15
[geometric_shape] Fix radius, CoM and inertia computation + Python unit tests

gabrielebndn:
`computeVolume()`, `computeCoM()` and `computeMomentofInertia()` had bugs, one of them really stupid concerning the volume of the sphere.
I fixed them and tested them in Python.
I suspect there is still a bug in the capsule inertia, which is why I have commented that part of the unit test, but I do not have time for that now

### DIFF
--- a/include/hpp/fcl/collision_object.h
+++ b/include/hpp/fcl/collision_object.h
@@ -125,10 +125,10 @@ public:
   FCL_REAL threshold_free;
 
   /// @brief compute center of mass
-  virtual Vec3f computeCOM() const { return Vec3f(); }
+  virtual Vec3f computeCOM() const { return Vec3f::Zero(); }
 
   /// @brief compute the inertia matrix, related to the origin
-  virtual Matrix3f computeMomentofInertia() const { return Matrix3f(); }
+  virtual Matrix3f computeMomentofInertia() const { return Matrix3f::Zero(); }
 
   /// @brief compute the volume
   virtual FCL_REAL computeVolume() const { return 0; }
--- a/include/hpp/fcl/collision_object.h
+++ b/include/hpp/fcl/collision_object.h
@@ -128,7 +128,7 @@ public:
   virtual Vec3f computeCOM() const { return Vec3f::Zero(); }
 
   /// @brief compute the inertia matrix, related to the origin
-  virtual Matrix3f computeMomentofInertia() const { return Matrix3f::Zero(); }
+  virtual Matrix3f computeMomentofInertia() const { return Matrix3f::Constant(NAN); }
 
   /// @brief compute the volume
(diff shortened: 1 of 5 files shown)


include/hpp/fcl/collision_object.h
```diff
@@ -125,10 +125,10 @@ class CollisionGeometry
   FCL_REAL threshold_free;
 
   /// @brief compute center of mass
-  virtual Vec3f computeCOM() const { return Vec3f(); }
+  virtual Vec3f computeCOM() const { return Vec3f::Zero(); }
 
   /// @brief compute the inertia matrix, related to the origin
-  virtual Matrix3f computeMomentofInertia() const { return Matrix3f(); }
+  virtual Matrix3f computeMomentofInertia() const { return Matrix3f::Zero(); }
```
jcarpent: I would suggest returning something which is not zero (NAN maybe), or having pure virtual methods.
I think this may lead to some bugs in the future.

gabrielebndn: No, this is how it should be. All subclasses inherit this and for most of them the correct result is zero. This is how it is in the original FCL itself. Before, they were not using Eigen and the class was self-initializing to zero. They there was the switch to Eigen, the self-initialization was not performed anymore and the result was garbage. They have the same fix in the original FCL

gabrielebndn: Oh, wait, you were commenting `computeMomentofInertia`... Yeah, it does not make sense for that. Still, I see no particular reason to change it

jcarpent: I don't know why this should be zero by default. Your answer is not convincing me, even if people have done that on the official FCL repository. Thinking it twice, may avoid future bugs I would say.

gabrielebndn: It makes sense for the CoM. It does not make a lot of sense as a default value for the moment of inertia. This is a fix, as far as I am concerned it comes "as is"

jcarpent: If you agree that this should not be the right value for the moment of inertia, then do not set this value. I mean, you should be coherent. Again, I'm trying to avoid future stupid bugs that will come later for sure.

jmirabel:
> I think the issue on computeMomentofInertia should be adressed before merging.

This piece of code isn't used anywhere. It is part of the prehistoric part of FCL, for which I never saw any use. I understand your concern and I am fine with any options.

gabrielebndn:
I confirm the capsule moment of inertia was wrong. I fixed it. The Python unit test mimics the derivation, you can check it

jcarpent:
@gabrielebndn Thanks for the fix. I really appreciate it. :)

PR #108, opened 2019-11-19
[CMake] export target hpp-fcl

jmirabel:


jmirabel:
I triggered a build on the build farm: https://gepgitlab.laas.fr/humanoid-path-planner/hpp-fcl/pipelines/6873

nim65s:
The buildfarm doesn't know yet that this package has python bindings, so they will not be tested for now. You can ask me to fix this if necessary.

jmirabel:
> The buildfarm doesn't know yet that this package has python bindings, so they will not be tested for now. You can ask me to fix this if necessary.

Not necessary as of today. My main concern is the finding and usage of the dependencies.

jmirabel:
All CI passed. This is ready for a release.
@nim65s can you take care of this ?

nim65s:
Perfect :)
I guess we can go for v1.2.0, or is there any non-backward-compatible changes that I missed ?

jmirabel:
A few, but I believe they are either not important or they concern an almost never used function.

Old API of `Transform3f`
- `inverse` does an in-place inversion (modifies the object)

New API  of `Transform3f`
- `inverseInPlace` does an in-place inversion (modifies the object)
- `inverse` returns the inverse of the object, without modifying the object (i.e. it is const)

Some functions were moved into namespace `hpp::fcl::internal` to make a clear distinction between what is the stable API.

nim65s:
Ok, this will do, thanks :)

jcarpent:
@nim65s Do you have some timings for the release. I can handle it if you want, together with Pinocchio and EigenPy. Is it OK for you?

nim65s:
The source release is already done, I'll push it here soon.

jcarpent:
Could you push it now? I need them to update all the conda-force and brew instances. I will do the release of Pinocchio after this step done.

nim65s:
done :)

jcarpent:
Thanks.

PR #112, opened 2019-12-04
cmake: refactor Python bindings to allow easy creation of wheels (pip)

jcarpent:
This refactorization won't change the behavior of the Python bindings, but will ease the creation of a wheel to do:
```
pip install hppfcl
```


jcarpent:
@jmirabel This PR also solves Travis check on OSX

jmirabel:
Thank you.

PR #114, opened 2019-12-05
Improve packaging + add ROS package

jcarpent:


src/CMakeLists.txt
```diff
@@ -111,8 +120,6 @@ IF(OCTOMAP_FOUND)
     -DOCTOMAP_PATCH_VERSION=${OCTOMAP_PATCH_VERSION})
 ENDIF(OCTOMAP_FOUND)
 
-target_include_directories(${LIBRARY_NAME} PUBLIC
-  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
```
jmirabel: You should not remove this. It adds to the exported target the include directory so that `target_link_libraries` can use it.

jmirabel: Ah, I see now that it was only moved.

jcarpent: Yes, I wanted to gather lines together ;).

jcarpent:
@jmirabel I think after Travis is happy, you can proceeed to the merge if you think it is fine.

PR #116, opened 2019-12-08
Improve bindings + remove borring warnings + fix

jcarpent:


include/hpp/fcl/math/transform.h
```diff
@@ -218,7 +219,7 @@ class Transform3f
   /// @brief check whether the transform is identity
   inline bool isIdentity() const
   {
-    return R.isIdentity() && T.isZero();
+    return R == Matrix3f::Identity() && T == Vec3f::Zero();
```
jmirabel: You change the semantic here. It may has some impact inside and outside hpp-fcl, so it cannot be changed without some care.

Moreover, I like the isIdentity which resembles Eigen's isIdentity method. May we should add an argument *threshold*.

What troubles you with the old behavior of  *isIdentity* ?

jcarpent: Yes, I changed the behavior intentionally to follow for instance the behavior of https://github.com/humanoid-path-planner/hpp-fcl/blob/e5c54ef26f1bfb4f189dc37c735c7655e38e961f/include/hpp/fcl/math/transform.h#L58
In addition, in some places inside HPP-FCL, some functions check that the transform is equal to Identity, but within a certain threshold, this means nothing (e.g. in the contect of finite differences for example.
This is why I changed the behavior, and the tests seem to pass.

@jmirabel Do you agree?

jmirabel: `isQuatIdentity`, `areQuatEquals` is declared and defined but never used.

> This is why I changed the behavior, and the tests seem to pass.

For hpp-fcl, this is not sufficient. The unit-tests are probably not good enough to detect possible numerical issues.

> some functions check that the transform is equal to Identity, but within a certain threshold, this means nothing

This may mean that there is a numerical issue when the transform is close to identity (shot in the dark). Note that this may happen in user code. In HPP, I think its use, if any, is limited to rbprm.

> e.g. in the contect of finite differences for example

I am not opposed to make hpp-fcl suitable for finite difference or automatic differentiation. However, this would be a big step that ought to be though globally.

> @jmirabel Do you agree?

I would prefer to discuss it in a separate issue or PR. The rest of this PR is straightforward.

jcarpent: I do understand. I've removed the commit.

jmirabel: Thank you.

PR #119, opened 2019-12-13
Capsule/Capsule distance function: reimplementation

rstrudel:


src/distance_capsule_capsule.cpp
```diff
@@ -25,299 +27,121 @@ namespace hpp
...
+    FCL_REAL halfLength1 = capsule1->halfLength;
+    FCL_REAL halfLength2 = capsule2->halfLength;
+    FCL_REAL radius1 = capsule1->radius;
+    FCL_REAL radius2 = capsule2->radius;
+    // direction of capsules
+    // ||d1|| = 2 * halfLength1
+    // Matrix3f::ConstColXpr d1 = 2 * halfLength1 * tf1.getRotation().col(2);
+    // Matrix3f::ConstColXpr d2 = 2 * halfLength2 * tf2.getRotation().col(2);
+    const fcl::Vec3f& d1 = 2 * halfLength1 * tf1.getRotation().col(2);
+    const fcl::Vec3f& d2 = 2 * halfLength2 * tf2.getRotation().col(2);
```
jmirabel: The ampersands should be removed.

rstrudel:
I encountered some bugs while trying the Capsule/Capsule collision detection. Additional tests have been written by Justin to check the collision function more thoroughly. The previous function did not pass these tests, thus I reimplemented the function using the book [Real-Time Collision Detection, Christer Ericson, p148-149](http://www.r-5.org/files/books/computers/algo-list/realtime-3d/Christer_Ericson-Real-Time_Collision_Detection-EN.pdf).

src/distance_capsule_capsule.cpp
```diff
@@ -25,299 +27,121 @@ namespace hpp
...
+    else if (e <= EPSILON)
+    {
+      // Second segment is degenerated
+      s = CLAMP(-c / a, 0.0, 1.0);
+      t = 0.0;
+    }
+    else
+    {
+      // Always non-negative, equal 0 if the segments are parallel
+      FCL_REAL denom = a*e-b*b;
```
jmirabel: Maybe the following would be more robust to numerical issues.
```
FCL_REAL denom = std::max(0, a*e-b*b);
```

src/distance_capsule_capsule.cpp
```diff
@@ -25,299 +27,121 @@ namespace hpp
...
+
+      if (denom != 0.0)
+      {
+        s = CLAMP((b*f-c*e) / denom, 0.0, 1.0);
```
jmirabel: CLAMP suffices here to make the division by denom robust to small values of denom, isn't it ?

jcarpent:
@jmirabel I will make a new release after this PR has been merged.

jcarpent:
I will put some comments here and @rstrudel is free to add his feedback.
@rstrudel realized that the current implementation of distance<Capsule, Capsule> was leading to incorrect behaviors when checking trivial situations as the ones reported in the unit tests. Currently, the previous implementation was not sufficiently tested in the tests. Maybe much more through the hpp pipeline.

Then, @rstrudel decided to implement the segment-to-segment distance check. Which seems appropriate for the Capsule-Capsule test, and in addition, leading to fewer codes. In addition, the reimplementation made by @rstrudel allows removing some constraints on the code (copy past from external lib).

jcarpent:
@rstrudel is currently using the collision between inside his RL framework. So, in a few days, we will see if there is still some bugs, hopefully not.

jmirabel:
> the previous implementation was not sufficiently tested in the tests.

definitely.

> leading to fewer codes.

I saw this and the code is a lot easier to understand.

jmirabel:
I double checked and I think capsule in HPP  are mostly used by @dbury.

jmirabel:
@DianeBury, do you have any comments on this topic ?

rstrudel:
Thanks for accepting the PR. @jcarpent summed it well, I am using hpp-fcl collision checker in my RL framework and I could observe bugs with capsules which lead to this PR. I will use it on a frequently in the following weeks and will see if it has the expected behavior.

PR #121, opened 2019-12-18
Default BUILD_PYTHON_INTERFACE to ON

gabrielebndn:
I think the bindings are sufficiently mature for building them by default. Do you agree with this?


jmirabel:
@nim65s can you make sure this is correctly handled by the HPP Makefile. It adds a dependency to eigenpy.

nim65s:
I would put this https://github.com/humanoid-path-planner/hpp-fcl/blob/ec0659566cb2e8810ebeb560c0db3eef0c3423c8/python/CMakeLists.txt#L35 into the main CMakeLists.txt, as I find it more explicit to list all potential dependencies in only one main place.

Apart from that, everything already works for me, so we can put this option to ON by default, yes.

gabrielebndn:
> I would put this
> 
> https://github.com/humanoid-path-planner/hpp-fcl/blob/ec0659566cb2e8810ebeb560c0db3eef0c3423c8/python/CMakeLists.txt#L35
> 
> into the main CMakeLists.txt, as I find it more explicit to list all potential dependencies in only one main place.
> Apart from that, everything already works for me, so we can put this option to ON by default, yes.

Makes sense. I'll do it right now

jmirabel:
Moving the dependency declaration is a good idea.

> Apart from that, everything already works for me, so we can put this option to ON by default, yes.

I think either
https://github.com/humanoid-path-planner/hpp-doc/blob/676b4d9f27e05053993dfa000fd1a170136955c6/doc/Makefile#L276
or
https://github.com/humanoid-path-planner/hpp-doc/blob/676b4d9f27e05053993dfa000fd1a170136955c6/doc/Makefile#L64
should be updated. I am in favor of the second.

gabrielebndn:
Done

gabrielebndn:
Hi! News on this PR? Can it be merged?

jmirabel:
For the devel branch, it is ok. The HPP Makefile still has to be updated.

PR #123, opened 2020-01-27
Expose additional quantities in Python

jcarpent:


jcarpent:
@rstrudel This might be interesting for you.

python/collision-geometries.cc
```diff
@@ -220,6 +227,76 @@ void exposeCollisionGeometries ()
...
+         "Check whether two AABB are overlaping and return the overloaping part if true.")
+  
+    .def("distance",
+         (FCL_REAL (AABB::*)(const AABB &) const)&AABB::distance,
+         bp::args("self","other"),
+         "Distance between two AABBs.")
+//    .def("distance",
+//         AABB_distance_proxy,
+//         bp::args("self","other"),
+//         "Distance between two AABBs.")
```
jmirabel: I guess AABB_distance_proxy is not necessary ?

jcarpent: Yes, because it is not implemented. I wanted to keep it for future, but I can also delete it. What do you request?

jmirabel: You want to bind the distance function which also returns the closest points ?
How to you want to disambiguate between the one without them and the one with them ?

jmirabel:
A note for later:

The following prototypes form the API common to all the bounding volumes. This may simplify binding them.

```c++
bool contain (const Vec3f& p) const;

bool overlap (const BV &other) const;
/// \param margin the BV overlap if the distance is inferior to this value.
bool overlap (const BV& other, const FCL_REAL& margin, FCL_REAL& sqrDistLowerBound) const;

FCL_REAL distance (const BV& other) const;
FCL_REAL distance (const BV& other, Vec3f& P, Vec3f& Q) const;

BV& operator+= (const Vec3f &p);
BV& operator+= (const BV& other);
BV operator+ (const BV& other) const;

FCL_REAL size () const;

const Vec3f& center () const;
FCL_REAL width () const;
FCL_REAL height () const;
FCL_REAL depth () const;
FCL_REAL volume () const;
```

Issue #106, opened 2019-11-16
Documentation of Python bindings


jcarpent:
I have tried the Python bindings, it seems to work well on my computer.
I did not try yet with Pinocchio, but I may give it a try next week.

In the meantime, I think it should be nice to provide the documentation with the main FCL geometries. @jmirabel Have you tried your script to automatize the documentation?

jmirabel:
I have something which works reasonably well, with the following drawbacks:
- it's a labyrinthine system that's gonna be hard to maintain (some kind of ro code...).
- handling CRTP will be very hard, if possible at all.

I will try to propose something in an (yet undefined) future.

jcarpent:
Thanks a lot for the quick reply. We can wait indeed. But, maybe we can put some manual documentation for standard primitives to help new users.

jmirabel:
I can add the generated files as a first step.

jmirabel:
@jcarpent , you may want to try the documentation on my branch python_doc.

You must add my fork of the JRL cmake module it order for it to work.

rstrudel:
Hi,

I do agree it would be nice to get more readable documentation. The typical output I get in python is:
`
In [4]: hppfcl.Sphere?                                                                                                                                                                                                                        
Docstring:      <no docstring>
Init docstring:
__init__( (object)arg1, (float)arg2) -> None :

    C++ signature :
        void __init__(_object*,double)
          ~/miniconda3/envs/robot/lib/python3.7/site-packages/hppfcl/hppfcl.cpython-37m-x86_64-linux-gnu.so
Type:           class
Subclasses
`

Whereas when I read the file exposing the shapes [collision-geometries.cc](https://github.com/humanoid-path-planner/hpp-fcl/blob/devel/python/collision-geometries.cc), it is well written and self explanatory. Sorry for the newbie/uninformed question, what is missing to get something that looks like the actual function signature in python ?

jmirabel:
I'll try to work on this.

rstrudel:
Thank you @jmirabel , I understand this is time consuming and that it is not really a priority. I have a limited knowledge of the repo and C++/python bindings in general but let me know if I can help at some point.

jmirabel:
@rstrudel A review of #129 is welcome.

jcarpent:
I've tried the generation of the documentation from doxygen on Mac, but it fails with the following error:
```
Error: /undefined in getenv
Operand stack:
   (outfile)
Execution stack:
[...]

PR #129, opened 2020-02-05
Generate Python documentation from doxygen documentation.

jmirabel:
Overall, this is interesting but has the few caveats:
- running `make` always triggers the generation of the documentation, which, in turn, always triggers compilation of the python bindings. In a development process, this is extremely annoying, since it slows down recompilation.
- I may have been a bit too ambitious. Handling template class and function turned out to be trickier than I thought.
- Some function may not be documented correctly (I don't remember the exact details but I think this is the case for the constructors of Transform)

Any ideas for improving this is welcome. Any help to complete Doxygen's documentation is also welcome, since it will also make the Python module documented.

Related to #106


jmirabel:
@jcarpent how do you install python-lxml on OSX ?

jmirabel:
I do not understand travis failure. Any feedback for someone trying to compile this is welcome ?

jmirabel:
> @jcarpent how do you install python-lxml on OSX ?

@jcarpent small up so that I can setup travis.

jmirabel:
Travis is fixed on linux. It is still failing because the script to generate the python doc uses python 3.

PR #133, opened 2020-02-21
Provide documentation for the Python bindings.

jmirabel:
Solves #106


python/fcl.cc
```diff
@@ -46,6 +46,12 @@
 
 #include <hpp/fcl/collision.h>
 
+#ifdef HAS_DOXYGEN_AUTODOC
```
jcarpent: Maybe the name of this flag may be adjusted with a prefix, to avoid naming clashes.

PR #137, opened 2020-03-10
Update submodule CMake and fix dependency to eigenpy

jcarpent:


jmirabel:
This seems to have broken the ros package. I am surprised that eigenpy is not >= 2.0 on ROS....

jcarpent:
Yes, I'm also surprised. @wxmerkt Do you know why eigenpy is still 1.6.9 on ROS?

PR #138, opened 2020-03-10
Minor fix and computational improvments

jcarpent:


src/narrowphase/details.h
```diff
@@ -261,15 +261,15 @@ namespace fcl {
...
+      const Vec3f & o2 = tf2.getTranslation();
       Vec3f diff = o1 - o2;
       FCL_REAL len = diff.norm();
       normal = -diff/len;
       dist = len - s1.radius - s2.radius;
 
-      p1 = o1 - diff * (s1.radius / len);
-      p2 = o2 + diff * (s2.radius / len);
+      p1.noalias() = o1 - normal * s1.radius;
+      p2.noalias() = o2 + normal * s2.radius;
```
jmirabel: If `normal = - diff / len`, then, shouldn't the sign before normal change ?

jmirabel: The normal points from s1 to s2 so the sign should be changed.

jcarpent: You're right. I have done the fix.

jcarpent:
@jmirabel You can merge when you want.

PR #140, opened 2020-03-13
Slight update of CMake

jcarpent:


python/CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ ELSE()
   ELSE()
     SET(ENABLE_DOXYGEN_AUTODOC FALSE)
     MESSAGE(STATUS "Python package lxml not found. Python bindings will not be documented.")
+    MESSAGE(STATUS "  You can install it with: pip install lxml")
```
jmirabel: Any reason to prefer the pip package other the apt one ?

nim65s: for debian / ubuntu, using the apt one is clearly better. But for OSX / Windows, pip is better :)

jcarpent: Mac users for instance.

PR #146, opened 2020-03-26
Improce code efficiency + use shared memory between Numpy and Eigen

jcarpent:
@jmirabel You've been merging your code while I was also doing some modifications. Hope the release of @nim65s is not already available.

This PR improves the code efficiency and also provides new way of sharing the memory between Eigen and Numpy.

### DIFF
--- a/include/hpp/fcl/narrowphase/narrowphase.h
+++ b/include/hpp/fcl/narrowphase/narrowphase.h
@@ -299,6 +299,8 @@ namespace fcl
     mutable Vec3f cached_guess;
   };
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wc99-extensions"
   /// \name Shape intersection specializations
   /// \{
 
@@ -403,6 +405,7 @@ namespace fcl
 #undef HPP_FCL_DECLARE_SHAPE_DISTANCE_PAIR
 
   /// \}
+#pragma GCC diagnostic pop
 }
 
 } // namespace hpp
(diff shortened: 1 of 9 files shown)


nim65s:
Don't worry for me :)

python/distance.cc
```diff
@@ -87,12 +91,10 @@ void exposeDistanceAPI ()
   if(!eigenpy::register_symbolic_link_to_registered_type<DistanceResult>())
   {
     class_ <DistanceResult> ("DistanceResult",
-        doxygen::class_doc<DistanceResult>(), init<>())
+                             doxygen::class_doc<DistanceResult>(),
+                             init<>(arg("self"),"Default constructor"))
```
jmirabel: FYI, this could also be
```c++
no_init)
.def (dv::init<DistanceResult>())
```
so the documentation (and the argument names) will follow the C++.

jcarpent: Done.

jmirabel:
Thanks for those updates.

include/hpp/fcl/narrowphase/narrowphase.h
```diff
@@ -299,6 +299,8 @@ namespace fcl
     mutable Vec3f cached_guess;
   };
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wc99-extensions"
```
jmirabel: @jcarpent I get a `warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]` because of this line. As this file is included in many places, the compilation output is completed flooded. Can you explain why you need this ?

Issue #147, opened 2020-03-27
Suspect bug in collide function on master and devel branches


gabrielebndn:
Hi,
while testing pinocchio on Ubuntu 19.04 I found that a unit test was not passing, and I suspect the root cause is in hpp-fcl.

So, in Pinocchio, I saw test [unittest/geom](https://github.com/stack-of-tasks/pinocchio/blob/master/unittest/geom.cpp) was not passing. This is the error I get:
```
/home/gabriele/devel/pinocchio/unittest/geom.cpp(100): error: in "geomTest/simple_boxes": check computeCollision(geomModel,geomData,0) == true has failed
/home/gabriele/devel/pinocchio/unittest/geom.cpp(112): error: in "geomTest/simple_boxes": check computeCollision(geomModel,geomData,0) == true has failed

*** 2 failures are detected in the test module "geomTest"
```

After struggling to understand why the github CI was passing, I realised I was not on the latest release, but on branch master, which is forward with respect to it. I can confirm the unit test passes on `v1.3.0`.

I suspect this is a bug in hpp-fcl, or that at least this is due to a non-backward-compatible change in hpp-fcl. What reinforces the suspect is that I sometimes don't get the same errors, suggesting an issue with uninitialized variables.

I investigated the Pinocchio unit test with valgrind, and indeed I found
```
Conditional jump or move depends on uninitialised value(s)
```

More precicely, after commenting everything but the first check:
```
==24940== Conditional jump or move depends on uninitialised value(s)
==24940==    at 0x4D154E5: hpp::fcl::DistanceResult::update(double, hpp::fcl::CollisionGeometry const*, hpp::fcl::CollisionGeometry const*, int, int, Eigen::Matrix<double, 3, 1, 0, 3, 1> const&, Eigen::Matrix<double, 3, 1, 0, 3, 1> const&, Eigen::Matrix<double, 3, 1, 0, 3, 1> const&) (collision_data.h:388)
==24940==    by 0x4D529A6: hpp::fcl::ShapeDistanceTraversalNode<hpp::fcl::Box, hpp::fcl::Box>::leafComputeDistance(int, int) const (traversal_node_shapes.h:150)
[...]

PR #149, opened 2020-04-01
Bug fix + prepare optimization of collision using GJK / EPA

jmirabel:
This fixes the two points raised in #147: 
- EPA failure on a pair of boxes,
- the uninitialized values when EPA fails.

The other commits are related to other changes.

### EPA failures
There is one EPA failure which I was not able to fix. It is due to the addition of parallel faces in the envelop built by EPA. This case should be handled in EPA::expand function.

### API of GJKSolver::shapeIntersect

Currently, a collision query between two shapes ends up in calculating the signed distance between the shapes. For pairs which use GJK / EPA to compute the signed distance, this is very costly because : 
- it isn't possible to break GJK as soon as it proves non-collision (or that a certain safety margin is ensured). This is especially useful for continuous shape.
- EPA is run even when not necessary.

This PRs fixes the old FCL API of shapeIntersect so that a distance lower bound can be computed. I won't have time to complete it now so I prefer to submit it as is. What is left for later is:
- make GJKSolver able to set the distanceEarlyBreak in GJK according to the requested security margin and break distance,
- modify the ShapeShapeCollide so that they use the ShapeCollisionTraversalNode.
Both changes should not be hard but they need to be carefully tested.

### DIFF
--- a/include/hpp/fcl/collision_data.h
+++ b/include/hpp/fcl/collision_data.h
@@ -223,9 +223,16 @@ public:
 
 public:
   CollisionResult()
+    : distance_lower_bound (std::numeric_limits<FCL_REAL>::max())
   {
   }
 
+  /// @brief Update the lower bound only if the distance in inferior.
+  inline void updateDistanceLowerBound (const FCL_REAL& distance_lower_bound_)
+  {
+    if (distance_lower_bound_ < distance_lower_bound)
+      distance_lower_bound = distance_lower_bound_;
+  }
 
   /// @brief add one contact into result structure
   inline void addContact(const Contact& c) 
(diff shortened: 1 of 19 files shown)


gabrielebndn:
Excellent! I cannot really judge the PR because I don't know enough, but I confirm it fixes the Pinocchio unit test in #147

jmirabel:
This also solves #151

nim65s:
What is the status of this PR ?
Should we include it in a new minor version, to fix https://github.com/stack-of-tasks/pinocchio/issues/1136 ?

jmirabel:
I have to solve the unit-test failure.

jmirabel:
Yes, we should include it in a new minor version.

jmirabel:
@nim65s from my side, this is ok. You can merge and release when CI is happy.

Issue #156, opened 2020-04-20
Build failure when building doxygen


wxmerkt:
I am building the latest devel of hpp-fcl and get the following error:

```
Traceback (most recent call last):
  File "/home/wxm/dev/install_ws/src/hpp-fcl/doc/python/doxygen_xml_parser.py", line 688, in <module>
    output = OutputStreams (args.output_directory, sys.stdout, sys.stderr))
  File "/home/wxm/dev/install_ws/src/hpp-fcl/doc/python/doxygen_xml_parser.py", line 531, in __init__
    self.tree = etree.parse (input)
  File "src/lxml/etree.pyx", line 3519, in lxml.etree.parse
  File "src/lxml/parser.pxi", line 1839, in lxml.etree._parseDocument
  File "src/lxml/parser.pxi", line 1865, in lxml.etree._parseDocumentFromURL
  File "src/lxml/parser.pxi", line 1769, in lxml.etree._parseDocFromFile
  File "src/lxml/parser.pxi", line 1163, in lxml.etree._BaseParser._parseDocFromFile
  File "src/lxml/parser.pxi", line 601, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 711, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 638, in lxml.etree._raiseParseError
IOError: Error reading file '/home/wxm/dev/install_ws/build/hpp-fcl/doc/doxygen-xml/index.xml': failed to load external entity "/home/wxm/dev/install_ws/build/hpp-fcl/doc/doxygen-xml/index.xml"
make[2]: *** [python/CMakeFiles/generate_doxygen_cpp_doc] Error 1
make[1]: *** [python/CMakeFiles/generate_doxygen_cpp_doc.dir/all] Error 2
make: *** [all] Error 2
```

The system is Ubuntu 18.04. Have you seen this before?

`/home/wxm/dev/install_ws/build/hpp-fcl/doc/doxygen-xml/index.xml` does not exist

jmirabel:
I have the same problem. I don't know why I missed it.
Thanks for the report. I will work on it.

jmirabel:
Should be solved now.

PR #157, opened 2020-04-20
Fix generation of XML documentation. 

jmirabel:
Solves #156


wxmerkt:
Thank you for the quick resolution!

PR #167, opened 2020-05-05
Add function to generate the convex hull.

jmirabel:
This PR enables to create a convex hull of a model at runtime. After some quick benchmarking, the performance of the convex-convex seems far better than BVH-BVH. 

This PR has some packaging pitfalls due to qhull packaging. I copy here some of the comment I wrote in the code.

```cmake
  # TODO We compile qhullcpp because it is not provided in the binary package while
  # the other parts of Qhull are released.
  # Compile Qhull package as found on github leads to a link error (because it generates
  # only a static library). One should be careful that the version of the qhull submodule
  # of this git repo matches the version of qhull provided by the system.
```

Any suggestion to improve this is welcome.


jmirabel:
> This PR has some packaging pitfalls due to qhull packaging.

To be more precise on what I meant: qhull contains 3 libraries (libqhull.(a/so), libqhull_r.(a/so), libqhullcpp.(a/so)). On 18.04, qhull furnishes the two first libraries and the headers of the 3 libraries. In this PR, I compile the part of qhull that generate libqhullcpp.(a/so) and link with the released two first libraries.

nim65s:
What about a robotpkg-libqhullcpp ?

jmirabel:
> What about a robotpkg-libqhullcpp ?

I didn't see it. I will check.

nim65s:
It doesn't exist yet, I think but I can create it.

jmirabel:
That would work for me. If you do it, there is one patch to apply in qhull CMakeLists.txt. The add_library(libqhullcpp) should be SHARED and not STATIC. I don't understand why I can't link to the STATIC lib though...

jcarpent:
Would it become a required dependency?

jmirabel:
> Would it become a required dependency?

At the moment, I would say no. If installing and managing qhull becomes easy, it would have nothing againt. The problem at the moment is that the git submodule must point to the same version as the installed one. This is the case on Ubuntu 18.04 but there is no reason this is true for other distrib so, at the moment, I don't consider it is easy to manage the qhull dependency.

jmirabel:
There are two work arounds:
- robotpkg (this is fine for me but I don't think it is sufficient to make the dependency required)
- using some CMake logic to download the sources corresponding to the installed version.

jcarpent:
Do you know why Qhull is only partially installed?

jmirabel:
> Do you know why Qhull is only partially installed?

No idea.

jmirabel:
One possible reason is that the two first libraries are old C libraries while the third one is more recent. Maybe the packaging in ubuntu was made before the third existed.

nim65s:
https://www.mail-archive.com/search?l=debian-bugs-dist@lists.debian.org&q=subject:%22Bug%23925540%5C%3A+qhull%5C%3A+add+libqhullcpp+to+installed+libraries%22&o=newest

jmirabel:
The patch in @nim65s second link is exactly what we need !

jmirabel:
Once CI is happy, can we merge it has it is now and come back to the packaging issue once we have a better handling of qhull ?

Meanwhile, @nim65s, on robotpkg, could you add qhull as a dependency (and set cmake flag HPP_FCL_HAS_QHULL to ON) for distributions where qhull is 7.2.0 ?
This is the case at least on 18.04 and 20.04.

jmirabel:
I am done with those changes.

jmirabel:
@nim65s is hpp-fcl compile with QHull on robotpkg ?

nim65s:
not yet

jcarpent:
With this respect, I think hpp-fcl should be linked statically to Qhull.

jcarpent:
There is a high-risk of incompatibility otherwise.

duburcqa:
> With this respect, I think hpp-fcl should be linked statically to Qhull.

I managed to link it statically. Is there any issue on your side ?

jcarpent:
What I meant here, is that the qhull used by @jmirabel should match the one on the system. It will be hard to have it strictly done in general. That's why, I think qhull should be packaged by FCL, and linked statically to the main lib. In such a way, there is no conflicts at all.

duburcqa:
And I totally agree on this point.

jmirabel:
> That's why, I think qhull should be packaged by FCL, and linked statically to the main lib. In such a way, there is no conflicts at all.

This is a possibility, yes. It has the advantage of solving the packaging issue. It may be a problem for people using both hpp-fcl and qhull in another package though.

@duburcqa , could you provide the CMake that links statically to the version of qhull within this package in a PR ?

PR #170, opened 2020-05-07
[Python] Remove unused headers.

jmirabel:


jmirabel:
@jcarpent FYI, hpp-fcl was not compiliing on 16.04 without this because of a weird error I could not understand. Just an info, no need to dig further.

jcarpent:
We need them to use the shared memory between Eigen and Numpy. This allows to export correctly any Eigen return type with reference. You should keep them.

jcarpent:
What is the error you got?

jmirabel:
I realized they were useful by seeing the unit test error. I tried to reproduce the bug in a clean docker and couldn't. I guess the issue comes from the computer where I had the issue.

PR #172, opened 2020-05-26
Fix packaging of the project

jcarpent:
At the time being, Boost components are not exported as a dependency via CMake packaging procedure. One way to fix this limitation is to use full lib path instead of targets.


jcarpent:
Afterwards, I will make a new release of the project. @nim65s @jmirabel Is it fine for you?

CMakeLists.txt
```diff
@@ -72,8 +72,8 @@ PROJECT(${PROJECT_NAME} ${PROJECT_ARGS})
 ADD_PROJECT_DEPENDENCY(Eigen3 REQUIRED PKG_CONFIG_REQUIRES "eigen3 >= 3.0.0")
 
 if(BUILD_PYTHON_INTERFACE)
-  ADD_PROJECT_DEPENDENCY(eigenpy 2.2 REQUIRED)
-endif ()
+  FIND_PACKAGE(eigenpy 2.2 REQUIRED)
```
nim65s, line 75: `ADD_PROJECT_DEPENDENCY` is a wrapper for `FIND_PACKAGE`, isn't it ? So why do you want to use `FIND_PACKAGE` directly ?

jcarpent: Because, eigenpy is only a build dependency.

nim65s: I thought we were installing hpp-fcl headers that were including eigenpy headers, but they are only included in sources. Thanks :)

jmirabel:
Any reason to change handling of assimp ?

jcarpent:
> Any reason to change handling of assimp ?

The whole idea is to be independent from pkg-config, which is missing from Windows for instance. And in the meantime, some assimp.pc on version 5.0.x are corrupted. I've made a fix, but some releases are not valid.

CMakeLists.txt
```diff
@@ -116,7 +120,7 @@ if(HPP_FCL_HAS_QHULL)
     )
 endif()
 
-ADD_REQUIRED_DEPENDENCY("assimp >= 2.0")
+FIND_PACKAGE(assimp REQUIRED)
```
jcarpent, line 123: @nim65s I think we should add `PKG_CONFIG_REQUIRES "assimp >= 2.0"`. No?

nim65s: I think assimp is only a build dependency, isn't it ? Would any hpp-fcl dependency have to include assimp headers, or link to assimp lib ?

Otherwise, this mean that we will have to put provide a way for dependencies to use the Findassimp.cmake, either by installing it along hpp-fcl-config.cmake, or by moving it into the jrl-cmakemodule.

jmirabel: @nim65s, you are right. It is internal only.

jcarpent: But you need it for the linkage procedure, if you only use pkg-config, you would like to have it mentioned.

jcarpent: @nim65s @jmirabel Do you agree?

jmirabel: If, using `find_package`, you are able to compile and link a package that depends on hpp-fcl, I think you don't need it in the pc file. Without it in the pc file, you will get the same flags as with the cmake config file.

I think the changes @nim65s proposes on linux. I don't know for other platforms.

nim65s: I don't understand why you would like to know that hpp-fcl depends on assimp for the linkage procedure of a package privately dependent on hpp-fcl but not using any assimp symbol.

jcarpent: My question is much more related to the linkage issue that might appear if you don't where assimp is located. But, maybe this is not needed. We can see and check afterwards.

nim65s: When you compile and link hpp-fcl, you do know where assimp is located thanks to the find_package. And when you compile and link a package that depends on hpp-fcl, I don't think you need to know anything about assimp.

I agree with you about the best way to find out: we can see and check afterwards. With explicit error messages we will be fixed.

src/CMakeLists.txt
```diff
@@ -173,28 +182,24 @@ endif()
 target_include_directories(${LIBRARY_NAME}
   SYSTEM PUBLIC
   ${EIGEN3_INCLUDE_DIR}
+  ${Boost_INCLUDE_DIRS}
+  ${assimp_INCLUDE_DIR}
   )
```
jmirabel, line 187: This one should not be public.
```suggestion
  )
target_include_directories(${LIBRARY_NAME}
  SYSTEM PRIVATE
  ${assimp_INCLUDE_DIR}
  )
```

nim65s: That's right, I did this for the library but I missed it for the include, thanks.

jmirabel: I let you take this change the way you like and merge this.

CMakeLists.txt
jcarpent: My worry was for people still relying on pkg config to link against Pinocchio or HPP in general.

jcarpent:
@nim65s Could you handle a new release to make Travis happy on Pinocchio?

PR #174, opened 2020-05-27
Compilation on Windows.

jmirabel:


jmirabel:
@nim65s I let you update the first commit to remove the spurious spaces.

Compilation works fine on my computer using flag `-fvisibility-ms-compat`.

CMakeLists.txt
```diff
@@ -41,10 +41,6 @@ set(PROJECT_DESCRIPTION
   )
 SET(PROJECT_USE_CMAKE_EXPORT TRUE)
 
-IF(NOT DEFINED CMAKE_CXX_STANDARD)
-  SET(CMAKE_CXX_STANDARD 98)
-ENDIF()
```
jmirabel, line 46: @nim65s I think we already discussed this but can't remember what we decided.
@jcarpent In a849427 (PR https://github.com/humanoid-path-planner/hpp-fcl/pull/116) you explicitly set the C++ standard. What was the reason ?

nim65s: In general, the good practice is not to set any standard. But in some places, we have to compile in C++ >= 11 even when the default compiler defaults to C++ 98. For this kind of cases, we currently force `CMAKE_CXX_STANDARD` to 11 in our CMakeLists or on robotpkg, but this is a bad idea.

I intend to provide a way to ensure that a minimum standard version is met in the JRL-cmakemodule, that will avoid setting any standard unless this is actually asked by the user or required by a dependency, and not compatible with the default of the current compiler.

But here, I really don't see the point for this `SET(CMAKE_CXX_STANDARD 98)`.

jcarpent: It was for removing borring warnings, you can remove it ;)

jcarpent:
I will add a CI on top of it. @jmirabel Did you check it on Windows?

jmirabel:
No. I don't know a simple way of testing Windows compilation. Do you have a docker image ?

jcarpent:
I don't know if docker is working with Windows. But, I'm trying to use now the Github Action service and on my machine, I've an a virtual box with Windows to test.

jmirabel:
> I don't know if docker is working with Windows.

https://stackoverflow.com/a/42185265 Windows container can only run on Windows.

jmirabel:
I just discover the existence of https://github.com/jrl-umi3218/github-actions. It supports Windows, but not its native compiler (instead it supports gcc and clang).

jcarpent:
I think the way it is done on EigenPy is the good one. It is based on Conda for installing dependencies, meaning that we don't have to recompile everything.

PR #176, opened 2020-06-03
Fix Python 3 doc generation

jmirabel:
This also bump the version in the package.xml.

I set it to the next release number because I think the devel branch should have a version number higher that the last release (except when they match, if course). On the practical aspect, this allows you to use stuff like `#ifdef HPP_FCL_VERSION_AT_LEAST(current devel version)` which will work with the next release.
Of course, we don't know yet whether next release will be 1.4.5 or 1.5 of 2.0 and we can adjust it upon release.

### DIFF
--- a/doc/python/xml_docstring.py
+++ b/doc/python/xml_docstring.py
@@ -69,10 +69,13 @@ class XmlDocString (object):
         if detailled is not None and len(detailled.getchildren()) > 0:
             if brief is not None: self._newline ()
             self.visit (detailled)
-        from sys import stdout, stderr
+        from sys import stdout, stderr, version_info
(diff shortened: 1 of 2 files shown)


nim65s:
Do you need `#ifdef HPP_FCL_VERSION_AT_LEAST(current devel version)` inside hpp-fcl, or in a dependent package ?

jmirabel:
In a dependent package.

jmirabel:
Inside hpp-fcl, you don't need it because the code is always synchronized.

nim65s:
Don't we have a `#ifdef HPP_FCL_GREATER_THAN(last released version)` ?

PR #177, opened 2020-06-07
Add Windows support and CI

jcarpent:


jmirabel:
Thank you for this @jcarpent . I looked at https://github.com/jcarpent/hpp-fcl/runs/747819654?check_suite_focus=true at, although the tick mark is green, the code hasn't actually been compiled. It stopped because CMake couldn't find assimp.

.github/workflows/windows-conda.yml
```diff
@@ -0,0 +1,58 @@
...
+        set BOOST_ROOT_1_72_0=
+        set PATH=%PATH:C:\hostedtoolcache\windows\Boost\1.72.0;=%
+
+        call "%programfiles(x86)%\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" amd64
+    
+        :: start building
+        mkdir build
+        pushd build
+
+        set PKG_CONFIG_PATH=%CONDA_PREFIX%\Library\share\pkgconfig:%CONDA_PREFIX%\Library\share\pkgconfig
```
jmirabel, line 46: I don't see any difference between the 2 paths here. Maybe this is a typo ?

jcarpent: Should be removed.

jcarpent:
The CI is now passing on Windows. Very good news.
But no more passing for the doc generation on Travis. Very bad news.

@jmirabel I let you manage this issue if you don't mind, because you know very well this feature.

jcarpent:
See https://github.com/jcarpent/hpp-fcl/actions for Windows CI.

jcarpent:
By the way, Octomap has not been exported for Windows. So I removed this feature for Windows.

jmirabel:
Doxygen is fooled by prototype like:
```cpp
HPP_FCL_DLLAPI void foo();
```
It only understands
```cpp
void foo() HPP_FCL_DLLAPI;
```
Is there a specific reason to put it in front ?

jcarpent:
It does not work otherwise :( The export symbol should be put forward.

jmirabel:
Ok, I will then fix it in the generation of the C++ doc then.

jmirabel:
For the reference: https://gcc.gnu.org/wiki/Visibility

jmirabel:
After double-checking, doxygen is fooled only by the macro because macro expansion is not enabled. It is very easy to fix.

jmirabel:
@jcarpent once CI is ok (which seems to be now), do you mind if I squash the commits of this PR into a single commit ?

jcarpent:
Keeping history will help us for debugging purposes in the future I would say. But if you think this is not clean, I don't really mind.

jmirabel:
Thank you @jcarpent for this PR !
@nim65s do you have time to release this ?

PR #181, opened 2020-06-23
Fix usage of Boost.

jmirabel:


CMakeLists.txt
```diff
@@ -76,14 +76,14 @@ if(BUILD_PYTHON_INTERFACE)
...
 # Required dependencies
-set(BOOST_COMPONENTS thread date_time system)
-set(BOOST_COMPONENTS ${BOOST_COMPONENTS} filesystem unit_test_framework chrono)
+ADD_PROJECT_DEPENDENCY(Boost REQUIRED)
+find_package(Boost REQUIRED COMPONENTS filesystem unit_test_framework chrono timer)
 if (BUILD_PYTHON_INTERFACE)
   FINDPYTHON()
-  set(BOOST_COMPONENTS ${BOOST_COMPONENTS} python)
+  set(BOOST_COMPONENTS python)
+  search_for_boost()
```
nim65s, line 84: I though you wanted a direct `search_for_boost_python()` macro. Did you changed your mind, or do you plan that for later ?

jmirabel: I didn't changed my mind on that. The job should be done in the CMake module first and it has a fairly low priority since the workaround is very simple.

jmirabel:
> I just hope that we'll not be dragged backwards by systems not yet compatible with `Boost::component` uses.

`Boost::unit_test_framework` was already there so I guess all systems we are interested in do support that.

PR #183, opened 2020-07-16
Add operator== for requests + bind StdVec of requests.

jmirabel:


python/distance.cc
```diff
@@ -89,6 +89,13 @@ void exposeDistanceAPI ()
       ;
   }
 
+  if(!eigenpy::register_symbolic_link_to_registered_type< std::vector<DistanceRequest> >())
```
jcarpent, line 92: We should do the same in Pinocchio ;)

jmirabel: If it is done here, why do you need it in Pinocchio ?

For consistency, I think it is better to define them all in the same package. I don't remember why I added it here. Maybe I hoped to remove some files in the Python bindings of Pinocchio.

jcarpent: It will depend on the ordering of the import in fact. Here, the cases Pinocchio + HPP-FCL will work well. No the other way.

jmirabel: In Pinocchio, the only binding related to hpp-fcl is the conversion between pinocchio::SE3 and hpp::fcl::Transform3f, isn't it ?

jcarpent: There is also the std::vector for Distance or Collision request if I remember well.

jmirabel: Yes, but they were all moved to hpp-fcl. In pinocchio, you have:
- `bp::import("hppfcl")` + the conversion SE3<->Transform3f in C++ and
- `from hppfcl import ...` in Python.
So this change will not affect Pinocchio at all.

jcarpent: Seems to be fine indeed.

jcarpent:
@jmirabel I've relaunch the CI and I will merge it when CI is green.
Afterwards, I will made a new release on top of #185.

jmirabel:
I think this last commit is required for the tests to pass.

jcarpent:
Perfect. Looks nice now.

PR #187, opened 2020-08-27
Logarithmic or linear support function for Convex depending of its size + Python update

jmirabel:


jcarpent:
I will provide a new release with the new packaging.

Issue #189, opened 2020-09-11
Impossible to compute collision between convex and halfspace


duburcqa:
- Trying to compute the collision between a convex geometry and an halfspace primitive through an exception
- The normal of the contact between a convex geometry and a box is always the one of the box, never the one of the convex geometry, no matter of their order in the collision pair. I'm not sure if it is intended and how to get the normal wrt the convex geometry.

NB: In Python, trying to access o1 from a CollisionResult raises an exception on my setup:
`No to_python (by-value) converter found for C++ type: hpp::fcl::CollisionGeometry const*`

duburcqa:
Also, do you know why I get such contact points / nearest points when computing the collision between a mesh geometry and a box ? 

In the figure, I'm computing the collision with the feet and a square box of size 2.0 centered and aligned with the origin:
![image](https://user-images.githubusercontent.com/17752950/92948805-4334f700-f45a-11ea-850d-2acd90f8c2df.png)

I would expect the contact points to be on the feet, but it is not the case. I don't have such issue when computing the collision with halfspace.

jcarpent:
Sorry but this issue is really unclear. Please clean it before we provide you an answer.

duburcqa:
@jcarpent Sure no problem. I edited it. It should be more clear.

florent-lamiraux:
Can you provide some sample code so that we can try to reproduce the issue ?

jmirabel:
> Trying to compute the collision between a convex geometry and an halfspace primitive through an exception

I never used this so I am not completely surprised that it throws something. I'll have a look quickly to see if it is easy to fix.

>     * The normal of the contact between a convex geometry and a box is always the one of the box, never the one of the convex geometry, no matter of their order in the collision pair.

*always* is surprising (depending on what your convex shape is). The normal is defined as the direction that yields the smallest translation to get out of collision. If you have a box which is big wrt the convex mesh, then, yes, I expect the normal to be one of the box normals.

>  I'm not sure if it is intended and how to get the normal wrt the convex geometry.

Not sure whether this makes sense because a) this normal does not exist when the most penetrating point is a vertex and b) does not yield a satisfactory direction.
What do you want to do exactly ?

jcarpent:
For the normal, I would say in the context of CONVEX vs BASIC PRIMITIVES, the solver should return a unique normal, which is easily defined for primitives, harder for meshes.

jmirabel:
[...]

PR #191, opened 2020-09-17
Add collision and distance between ConvexBase and Halfspace. 

jmirabel:
Partially solves #189 

I haven't tested this. @duburcqa, you may be able to test it quickly in your setup.


src/narrowphase/details.h
```diff
@@ -1942,6 +1942,29 @@ namespace fcl {
...
+                                  const ShapeBase& s, const Transform3f& tf2,
+                                  FCL_REAL& dist, Vec3f& p1, Vec3f& p2,
+                                  Vec3f& normal)
+    {
+      Vec3f n_w = tf1.getRotation() * h.n;
+      Vec3f n_2 (tf2.getRotation().transpose() * n_w);
+      int hint = 0;
+      p2 = getSupport(&s, -n_2, true, hint);
+      dist = p2.dot(n_2) - h.d;
+      p1 = p2 - dist * n_2;
```
jcarpent, line 1959: Maybe a noalias here can be used.

src/distance/triangle_halfspace.cpp
```diff
@@ -0,0 +1,81 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2011-2014, Willow Garage, Inc.
```
jcarpent, line 4: About licensing here, I do think only the copyright of CNRS should appear here because this is new content.

src/narrowphase/details.h
jmirabel: There is no possible aliasing as dist is a scalar.

jcarpent (approved):
I made just minor remarks.

src/narrowphase/details.h
jcarpent: Yes, the you can directly use noalias to avoid intermediate allocation (even on the stack).

src/distance/triangle_halfspace.cpp
jmirabel: Done.

duburcqa:
Nice thank you ! I'm going to test it asap ! Hopefully tomorrow

python/collision.cc
```diff
@@ -115,8 +120,12 @@ void exposeCollisionAPI ()
         doxygen::class_doc<Contact>(), init<>(arg("self"),"Default constructor"))
       //.def(init<CollisionGeometryPtr_t, CollisionGeometryPtr_t, int, int>())
       //.def(init<CollisionGeometryPtr_t, CollisionGeometryPtr_t, int, int, Vec3f, Vec3f, FCL_REAL>())
-      .DEF_RO_CLASS_ATTRIB (Contact, o1)
-      .DEF_RO_CLASS_ATTRIB (Contact, o2)
+      .add_property ("o1",
+          make_function(&geto<1>, return_internal_reference<>()), &seto<1>,
```
jcarpent, line 124: I don't think this would work well. Because, o1 and o2 might be null and there are not managed by Collision

jcarpent (changes requested):

python/collision.cc
jmirabel: I updated the code following our private discussion. Is this correct now ?

jmirabel:
> Nice thank you ! I'm going to test it asap ! Hopefully tomorrow

If you can make a decent self-contained Python example, that would be wonderful.

python/collision.cc
jcarpent: You are still missing the check whether the pointer is empty ;)

jmirabel: I tried the above code and I get a None when the pointer is NULL. Do I really have to handle it manually ?

jcarpent: I will try on my side but seems to be fine. After merging this PR, I will make a new release. It would also be nice to add testing over the Python bindings to avoid future bugs.

jmirabel: > It would also be nice to add testing over the Python bindings to avoid future bugs.

Yes, that would be nice. We also need to unit-test what I added in this PR. I will wait for @duburcqa feedback.

jcarpent: Sounds good to me.

jcarpent (approved):

jcarpent:
@duburcqa We need your test in order to merge and provide an new release. When could you provite it?

jmirabel:
There is now a very minimalist unit test.

duburcqa:
Are you sure you added it ? Mine would be minimalist too because it is pretty difficult for me to make a unit test out of the whole programm requiring this type of collision.

jmirabel:
Indeed, I forgot a file.

jcarpent:
I will merge it now and provide a new release then

PR #194, opened 2020-09-30
Add DistanceFunctor

jmirabel:
This is a draft to open discussions on removing the overhead introduced by hpp-fcl design.

The following benchmark for capsule-capsule distance shows that a hpp-fcl design introduces about 12ns per call, which is about 23% of the current cost. The attempt to remove this cost proposed here makes the overhead drop to 2ns, so less than 5%.

**NOTE** this figures should be taken with care.

```c++
#include <hpp/fcl/shape/geometric_shapes.h>
#include <hpp/fcl/distance.h>
#include "../src/distance_func_matrix.h"
#include "utility.h"

#include <iostream>
#include <chrono>

template<typename TimeT = std::chrono::milliseconds>
struct measure
{
    template<typename F, typename ...Args>
    static typename TimeT::rep execution(F&& func, Args&&... args)
    {
        auto start = std::chrono::steady_clock::now();
        std::forward<decltype(func)>(func)(std::forward<Args>(args)...);
        auto duration = std::chrono::duration_cast< TimeT>
                            (std::chrono::steady_clock::now() - start);
        return duration.count();
    }
};

using namespace hpp::fcl;

std::vector<Transform3f> transforms;
Transform3f Id;

Capsule c1 (0.2, 1.),
        c2 (0.12, 0.7);

DistanceRequest request (true);

void with_hpp_fcl_overhead ()
{
  for (auto M : transforms) {
    DistanceResult result;
    distance(&c1, Id, &c2, M, request, result);
  }
}

void with_minimal_hpp_fcl_overhead ()
{
  DistanceFunctor func (&c1, &c2);
  for (auto M : transforms) {
    DistanceResult result;
    func.distance(Id, M, request, result);
  }
}

void without_hpp_fcl_overhead ()
{
  for (auto M : transforms) {
    DistanceResult result;
    ShapeShapeDistance<Capsule, Capsule>(&c1, Id, &c2, M, NULL, request, result);
  }
}

int main ()
{
  FCL_REAL extends[] = { 2, 2, 2, 2, 2, 2 };
  int n = 1e7;
  generateRandomTransforms(extends, transforms, n);

  using std::chrono::nanoseconds;
  std::cout << "current  : " << measure<nanoseconds>::execution(with_hpp_fcl_overhead            ) / n << std::endl;
  std::cout << "new      : " << measure<nanoseconds>::execution(with_minimal_hpp_fcl_overhead    ) / n << std::endl;
  std::cout << "smallest : " << measure<nanoseconds>::execution(without_hpp_fcl_overhead         ) / n << std::endl;

  return 0;
}
```
gives something like
```
current  : 60
new      : 48
smallest : 46
```

### DIFF
--- a/src/distance.cpp
+++ b/src/distance.cpp
@@ -119,7 +119,32 @@ FCL_REAL distance(const CollisionGeometry* o1, const Transform3f& tf1,
   return res;
 }
 
+ComputeDistance::ComputeDistance(const CollisionGeometry* o1,
+    const CollisionGeometry* o2)
+  : o1(o1), o2(o2)
+{
+  const DistanceFunctionMatrix& looktable = getDistanceFunctionLookTable();
 
+  OBJECT_TYPE object_type1 = o1->getObjectType();
+  NODE_TYPE node_type1 = o1->getNodeType();
+  OBJECT_TYPE object_type2 = o2->getObjectType();
+  NODE_TYPE node_type2 = o2->getNodeType();
+
+  swap_geoms = object_type1 == OT_GEOM && object_type2 == OT_BVH;
+
+  if(   ( swap_geoms && !looktable.distance_matrix[node_type2][node_type1])
+     || (!swap_geoms && !looktable.distance_matrix[node_type1][node_type2]))
+  {
+    std::ostringstream oss;
+    oss << "Warning: distance function between node type " << node_type1 <<
+      " and node type " << node_type2 << " is not supported";
+    throw std::invalid_argument(oss.str());
+  }
+  if (swap_geoms)
+    func = looktable.distance_matrix[node_type2][node_type1];
+  else
+    func = looktable.distance_matrix[node_type1][node_type2];
 }
 
+} // namespace fcl
 } // namespace hpp
--- a/include/hpp/fcl/collision.h
+++ b/include/hpp/fcl/collision.h
@@ -109,7 +109,7 @@ public:
     std::size_t res;
     if (swap_geoms) {
       res = func(o2, tf2, o1, tf1, &solver, request, result);
-      invertResults(result);
+      result.swapObjects();
     } else {
       res = func (o1, tf1, o2, tf2, &solver, request, result);
     }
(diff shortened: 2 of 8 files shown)


src/distance.cpp
```diff
@@ -119,6 +119,42 @@ FCL_REAL distance(const CollisionGeometry* o1, const Transform3f& tf1,
...
+  : o1(o1), o2(o2)
+{
+  const DistanceFunctionMatrix& looktable = getDistanceFunctionLookTable();
+
+  OBJECT_TYPE object_type1 = o1->getObjectType();
+  NODE_TYPE node_type1 = o1->getNodeType();
+  OBJECT_TYPE object_type2 = o2->getObjectType();
+  NODE_TYPE node_type2 = o2->getNodeType();
+
+  if(object_type1 == OT_GEOM && object_type2 == OT_BVH)
```
jcarpent, line 133: The if/then/else struct should be revisited to avoid extra codes and provide simplifications.

jcarpent (approved):
Seems to be very relevant for basic shapes indeed. The impact should be less impressive on more complex geometries. But, I'm totally in favour.

jmirabel:
With 4e39e2f , I get
```
current  : 56
new      : 42
smallest : 41
```

jmirabel:
I did the same thing for collision queries.

jcarpent:
For Windows, you need to add the `HPP_FCL_DLLAPI` for making it working

jmirabel:
This PR is ready. I am not very satisfied with `CollisionFunctor` and `DistanceFunctor`. Any ideas for a better name ?
I can think of `CollisionPair`.

florent-lamiraux:
> This PR is ready. I am not very satisfied with `CollisionFunctor` and `DistanceFunctor`. Any ideas for a better name ?
> I can think of `CollisionPair`.

`ComputeCollision`, `ComputeDistance` would result in code like:
```
ComputeCollision cc(obj1, obj2);
cc(tf1, tf2, request, result);
```

jcarpent:
I like the suggestion of @florent-lamiraux. I think this is self-content and clear enough.

jmirabel:
I took @florent-lamiraux suggestion.

PR #196, opened 2020-10-03
Add testing on Windows for v142 compiler

jcarpent:


jcarpent:
I've merged. I will update the CMake file and provide a new release 1.6.0.

PR #198, opened 2020-10-23
Fix potential bug when returning depth distance when computing collision

jcarpent:


src/collision_func_matrix.cpp
```diff
@@ -71,44 +72,33 @@ std::size_t Collide(const CollisionGeometry* o1, const Transform3f& tf1, const C
...
   if(request.isSatisfied(result)) return result.numContacts();
 
   DistanceResult distanceResult;
   DistanceRequest distanceRequest (request.enable_contact);
-  FCL_REAL distance = ShapeShapeDistance <T_SH1, T_SH2>
+  const FCL_REAL distance = ShapeShapeDistance <T_SH1, T_SH2>
     (o1, tf1, o2, tf2, nsolver, distanceRequest, distanceResult);
 
-  if (distance <= 0) {
+  if (distance <= 0 || distance <= request.security_margin) {
```
florent-lamiraux, line 87: I would remove the test `distance <= 0`

src/collision_func_matrix.cpp
```diff
@@ -71,44 +72,33 @@ std::size_t Collide(const CollisionGeometry* o1, const Transform3f& tf1, const C
...
     if (result.numContacts () < request.num_max_contacts) {
       const Vec3f& p1 = distanceResult.nearest_points [0];
       const Vec3f& p2 = distanceResult.nearest_points [1];
+      const Vec3f& normal = distance <= 0 ? distanceResult.normal : (p2-p1).normalized();
```
florent-lamiraux, line 91: if `distance <= 0`, `p1` is likely to be equal to `p2`.

src/collision_func_matrix.cpp
```diff
@@ -71,44 +72,33 @@ std::size_t Collide(const CollisionGeometry* o1, const Transform3f& tf1, const C
...
 
       Contact contact (o1, o2, distanceResult.b1, distanceResult.b2,
-          (p1+p2)/2,
-          distanceResult.normal,
-          -distance+request.security_margin);
+                       (p1+p2)/2,
+                       normal,
+                       -distance);
```
florent-lamiraux, line 96: I think the penetration should take into account the security margin. Volumes should be thought of as if they were inflated by half the security margin. In any case, It must be consistent with other parts of the code.

florent-lamiraux (commented):
Have tested or encountered issues with this part of the code ?

src/collision_func_matrix.cpp
jcarpent: I don't know if I would remove it because users may provide positive security_margin. No?

src/collision_func_matrix.cpp
jcarpent: Not really, because it means you have interpenetration.

jcarpent:
> Have you tested these modifiations ? Have you encountered issues with this part of the code ?

I've discussed this PR with @jmirabel on personal messages. The thing is that, at the current stage, if:
- two objects interpenetrate, then the returned depth_penetration is positive, and is equal to `-distance + security_margin`
- the distance between the objects is greater than the security_margin, then there is no collision (as expected).
- if the two objects are not in collision, but their distance is lower than security_margin, then the two objects are considred in collision and the depth_penetration is equal to `-distance + security_margin`, meaning that depth_penetration remains positive.

I would rather imagine that the depth_penetration should be a signed value:
- positive, when two objects have a common intersection, with value equal to 0 when the objects are just contacting.
- negative, when the two objects do not intersect, but are at a distance <= security_margin
And there is no collision considered otherwise.

src/collision_func_matrix.cpp
jcarpent: I do not agree on this aspect. The `penetration_depth` should be a signed value. In the current context, this is not the case. This PR is trying to fix it.

florent-lamiraux:
I have no preference on the definition of the penetration depth. It should be consistent for all pairs of shapes. This implies to modify all the specific computations and there are many of them. I have pushed a test on branch "devel" of git@github.com:florent-lamiraux/hpp-fcl.git that confirms the current behavior as expressed above for 2 spheres.

jcarpent:
I do not see other places where the code should be adjusted if we agree on this convention. Could you provide me some pointers?

jmirabel (changes requested):
The definition we should stick to are:
image

With a security margin, what is currently implemented is
image

Function `collision` adds a `Contact` object when `signed_distance_margin(A,B) <= 0`. Class `Contact` has a field for storing the penetration depth (thus unsigned) and this make sense because there is no contact if the signed distance is positive.

@jcarpent, re-reading our discussion, why don't you should use function `distance` rather than `collision` ? I think you would have what you want.

Before we continue on the code, we should agree on the API, if the current one does not fulfill the needs.

Then, if we modify the API, as @florent-lamiraux suggested, there are other places to update (e.g. the BVH <-> something collision functions).

jmirabel:
After double thinking about it, I think what @jcarpent requests (tell me if I am wrong) is not implemented but the request make sense and shouldn't be hard to implement.

@jcarpent wants (again tell me if I am wrong):
- the contact information as they are now provided when the objects are in contact,
- the closest points when `0 <= signed_distance_margin(A,B) <= threshold`.

The security margin is not the correct way of handling this. I think the correct way of handling this is the request attribute `break_distance`. At the moment, when `0 <= signed_distance_margin(A,B) <= break_distance`, then the `distance_lower_bound` contains `signed_distance_margin(A,B)`. The only thing which is left to be done is to retain the closest points.

Would we agree on this ?

jcarpent:
Yes, seems to be a very good solution.

jmirabel:
I close this in favor of #200

Issue #199, opened 2020-10-26
Method impossible to call from Python


duburcqa:
Hi,

AFAIK, `CollisionObject` is not exposed in Python, so it is impossible to call this method: https://github.com/humanoid-path-planner/hpp-fcl/blob/e63c349201297ef46ef651ca1e7883a7d4ab2571/python/distance.cc#L129

If I'm right, I suggest either to expose `CollisionObject` or to remove it ? The same goes for collisions.

Best

jmirabel:
You are right, this overload isn't accessible.
I am not sure of the benefit of CollisionObject but since it exists in c++, it could be exposed in Python.

duburcqa:
I must admit I like it because in practice for my application one of the geometry is fixed, so it avoids me passing systematically the same transform (even though it is the same internally). And to my view, it is closing the Pinocchio way of thinking (aka Model/Data).

duburcqa:
Also since I'm here, I would like to point out the fact that `fcl::Octree` are not exposed in Python 

https://github.com/humanoid-path-planner/hpp-fcl/blob/a140097eb8ba171a807ac76009747a3e356fc93f/python/collision-geometries.cc#L416

So they appears as `CollisionGeometry`, making impossible to call some useful Octree-specific methods, especially `toBoxes`. But it is another issue not related to this one.

jmirabel:
Are you wiling to propose a PR exposing CollisionObject and Octree ?

duburcqa:
Not really for now. I'm not using hppfcl in Python. I was just trying to do some quick prototyping when I noticed those limitations.

jmirabel:
CollisionObject has been exposed in #212

PR #200, opened 2020-10-27
Add CollisionResult::nearest_points

jmirabel:
This supersedes #198 .

Work left to be done:
- [ ] agree on the API
- [x] handle shape-shape collision
- [ ] add some unit test
- [x] revise the distance lower bound (does not take the security margin).

Class ShapeCollisionTraversal should be removed.

As a bonus, this also solves #199.

### DIFF
--- a/include/hpp/fcl/internal/traversal_node_bvh_shape.h
+++ b/include/hpp/fcl/internal/traversal_node_bvh_shape.h
@@ -196,10 +196,8 @@ public:
       disjoint = !overlap(this->tf1.getRotation(), this->tf1.getTranslation(),
                       this->model2_bv, this->model1->getBV(b1).bv,
                       this->request, sqrDistLowerBound);
-    if (   disjoint
-        && this->result->distance_lower_bound > 0
-        && sqrDistLowerBound < this->result->distance_lower_bound * this->result->distance_lower_bound)
-      this->result->distance_lower_bound = sqrt(sqrDistLowerBound);
+    if (disjoint)
+      internal::updateDistanceLowerBoundFromBV(this->request, *this->result, sqrDistLowerBound);
     assert (!disjoint || sqrDistLowerBound > 0);
     return disjoint;
   }
@@ -255,11 +253,9 @@ public:
       }
     } else
       sqrDistLowerBound = distance * distance;
-    if (distToCollision < this->result->distance_lower_bound) {
-      this->result->distance_lower_bound = distToCollision;
-      this->result->nearest_points[0] = c1;
-      this->result->nearest_points[1] = c2;
-    }
+
+    internal::updateDistanceLowerBoundFromLeaf(this->request,
+        *this->result, distToCollision, c1, c2);
 
     assert (this->result->isCollision () || sqrDistLowerBound > 0);
   }
@@ -314,10 +310,8 @@ public:
       disjoint = !overlap(this->tf2.getRotation(), this->tf2.getTranslation(),
                      this->model1_bv, this->model2->getBV(b2).bv,
                      sqrDistLowerBound);
-    if (disjoint
-        && this->result->distance_lower_bound > 0
-        && sqrDistLowerBound < this->result->distance_lower_bound * this->result->distance_lower_bound)
-      this->result->distance_lower_bound = sqrt(sqrDistLowerBound);
+    if (disjoint)
+      internal::updateDistanceLowerBoundFromBV(this->request, *this->result, sqrDistLowerBound);
     assert (!disjoint || sqrDistLowerBound > 0);
     return disjoint;
   }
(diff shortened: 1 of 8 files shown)


jmirabel:
@nim65s does it make sense to turn off travis (except for osx) ?

nim65s:
I prefer not to have opinions about travis. If someone wants to use and maintain it, that's perfectly fine for me, but on my side I have enough work and results with one CI stack :)

jmirabel:
After some effort, I ended up with the following.
- for collision queries, all the returned distances take the security margin into account,
- the collision result stores valid nearest_points whenever the distance lower bound is inferior to `break_distance`. When the distance lower bound is higher, their validity depends upon the pair of object type (maybe we should add a boolean that says whether they are valid).
- the distance lower bound can be negative, in which case it is the opposite of the penetration depth (the security margin is taken into account).

This leads to one strange thing. With the current changes, when the distance lower bound is inferior to the break distance, we don't have `||AB|| = distance_lower_bound` but ||AB|| = distance_lower_bound + security_margin` (A and B being the nearest point). i.e. the nearest point are not on the shapes inflated by `security_margin/2`

Some improvements, kept for later:
- For BVH, the minimal distance between two bounding volumes could be decrease as the algo finds smaller distance lower bound. This is especially useful when the `break_distance` is set to a high value.
- Rename `break_distance` to something easier to understand from the user perspective.
- Clean API of CollisionTraversalNode:
  - remove ShapeCollisionTraversalNode,
  - merge the two CollisionTraversalNode::BVDisjoint prototype into one. The squared distance lower bound attribute is not useless.
- The above API should be enforced by some unit-tests.

jmirabel:
This PR is ready for reviews.

src/collision.cpp
```diff
@@ -61,6 +61,7 @@ void CollisionResult::swapObjects()
   {
     std::swap(it->o1, it->o2);
     std::swap(it->b1, it->b2);
+    it->normal *= -1;
```
jcarpent, line 64: This is a critical bug, isn't ?

jcarpent (approved):

src/collision.cpp
jmirabel: Yes. I am surprised it hasn't popup earlier.

include/hpp/fcl/internal/traversal_node_bvh_shape.h
```diff
@@ -337,27 +346,32 @@ class ShapeMeshCollisionTraversalNode : public ShapeBVHCollisionTraversalNode<S,
...
-    }
-    assert (!this->result->isCollision () || sqrDistLowerBound > 0);
+    } else if (distToCollision < 0) {
+      sqrDistLowerBound = 0;
+      if(this->request.num_max_contacts > this->result->numContacts())
+      {
+        this->result->addContact(Contact(this->model1, this->model2,
+              Contact::NONE, primitive_id,
+              .5 * (c1+c2), (c2-c1).normalized (),
+              -distance));
```
jmirabel, line 366: TODO: sometimes the contact distance takes the security margin into account. Sometimes it does not. It should be homogenized.

src/collision_func_matrix.cpp
```diff
@@ -82,34 +82,20 @@ std::size_t ShapeShapeCollide(const CollisionGeometry* o1, const Transform3f& tf
...
+  const Vec3f& p1 = distanceResult.nearest_points [0];
+  const Vec3f& p2 = distanceResult.nearest_points [1];
+  FCL_REAL distToCollision = distance - request.security_margin;
 
-      Contact contact (o1, o2, distanceResult.b1, distanceResult.b2,
-          (p1+p2)/2,
-          distanceResult.normal,
-          -distance+request.security_margin);
+  internal::updateDistanceLowerBoundFromLeaf (request, result, distToCollision, p1, p2);
+  if (distToCollision <= 0 && result.numContacts () < request.num_max_contacts) {
```
jcarpent, line 90: Currently, this results is not taking into account the break_distance information.
Compared to other collision checks.
```suggestion
  if (distToCollision <= request.break_distance && result.numContacts () < request.num_max_contacts) {
```

jcarpent (changes requested):
Maybe, this should fix another bug

src/collision_func_matrix.cpp
jmirabel: No, there is a contact only if distToCollision <= 0. At the moment, `break_distance` is not useful for shape-shape pairs.

include/hpp/fcl/internal/traversal_node_bvh_shape.h
jcarpent: Why to not apply the change here?

jmirabel: I plan to do it. I wrote it so as not to forget.

jcarpent:
@jmirabel Could you consider adjusting this PR for merging it?

jmirabel:
Not sure what you are asking exactly. I can definitely resolve the git conflicts quickly. Adding unit test, finalizing the code and maybe update the API is another story. I know I won't have time before beginning of March. Then I am not sure yet.

jmirabel:
I assigned myself so that I see this in my list of assigned PRs.

src/collision_node.cpp
```diff
@@ -60,7 +60,13 @@ void collide(CollisionTraversalNodeBase* node,
       collisionRecurse(node, 0, 0, front_list, sqrDistLowerBound);
     else
       collisionNonRecurse(node, front_list, sqrDistLowerBound);
-    result.updateDistanceLowerBound (sqrt (sqrDistLowerBound));
```
jcarpent, line 63: @jmirabel Why do you remove that here?

jcarpent:
Supersedes by #303.

PR #203, opened 2020-11-05
[CI] Disable travis on linux (duplicates gitlab CI).

jmirabel:
Will help regarding travis policy changes : https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing


jcarpent:
The thing is the CI on gitlab may fail just because the containers are not up to date, and so on.
Maybe, we can move the Travis scripts to Github Action.

jmirabel:
@nim65s any comment on the reliability of the CI on gitlab ?

nim65s:
I don't think we ever got any Gitlab-CI issue with this package. Did we ? 

In any case, I am not aware of any ongoing issue. Should I ?

jcarpent:
On my case, I often encounter for Pinocchio at least the case where some runners are stuck and then I add to manually relaunch the build, which is boring in some sense.
On the other side, it would be nice to keep checking for OS X thanks to Gitlab.

jmirabel:
>  On the other side, it would be nice to keep checking for OS X thanks to Gitlab.

What annoys me the most with the current CI on OSX is the fact that most of the time is spent compiling dependencies...

jcarpent:
This is due to the very old version OSX on travis.
One solution would be using conda, as for Windows.

nim65s:
> On my case, I often encounter for Pinocchio at least the case where some runners are stuck and then I add to manually relaunch the build, which is boring in some sense.

This can probably easily be workarounded. For now, it is configured to retry twice on `runner_system_failure`, but there are [many other options](https://docs.gitlab.com/ee/ci/yaml/#retry) we can consider.

Please, ping me the next time you need to manually retry a job :)

jmirabel:
Can catalina be put inside a docker ?

nim65s:
Technically, yes. Legally… That's a grey zone.

jmirabel:
So how do you do CI on mac ? You have to buy a license ?

nim65s:
The easiest way is to have runner directly on macOS on an Apple computer. Same for windows: https://docs.gitlab.com/runner/install/osx.html

jcarpent:
OS X is free.

jcarpent:
We can run OS X on GitHub Action. The remaining question is how to install the binaries of the dependencies. One of the quick answers is to use conda, which provide binaries for Linux, OS X and Windows.

jmirabel:
> We can run OS X on GitHub Action. The remaining question is how to install the binaries of the dependencies. One of the quick answers is to use conda, which provide binaries for Linux, OS X and Windows.

@jcarpent I think I will disable OSX on travis. Would you have time to setup CI on github using conda ?

jcarpent:
Later in the year.

jcarpent:
Duplicated with #229.

PR #204, opened 2020-11-12
Fix doc generation for Python bindings on Windows

jcarpent:


doc/python/doxygen_xml_parser.py
```diff
@@ -408,7 +409,7 @@ def _writeClassDoc (self, output):
             classname = self._className(),
             docstring = docstring,
             attributes = attribute_docstrings,
-            ))
+            ).encode('utf-8'))
```
nim65s, line 412: You can specify the encoding in the [`open`](https://docs.python.org/3/library/functions.html#open) function of the file, instead of in all the streams you are giving to that file, ie. on l. 652:
```python
self._out = open(fullname, mode='w', encoding='utf-8')
```

jcarpent: Since which version? 
I also had some issues sayong that the str was not compatible when performing testing.

jcarpent: I removed it again. let's see if it works.

nim65s: Good catch, that's not available in python 2… 

We could add it in the `out()` method, l.675, ie:
```python
def out(self, *args):
    print(' '.join(str(arg) for arg in args).encode('utf-8'), file=self._out)
```

But that's not as clean. I let you choose :)

jcarpent: I think the doc generation is only for Python3 ;)

nim65s: then, the `encoding` arg of the `open()` function is available in all python 3.x versions

jcarpent: Solved.

PR #205, opened 2020-11-15
BVH - Add new methods + fix some bugs + remove some warnings

jcarpent:


include/hpp/fcl/internal/traversal_node_setup.h
```diff
@@ -308,16 +308,16 @@ bool initialize(MeshShapeCollisionTraversalNode<BV, S>& node,
                 bool use_refit = false, bool refit_bottomup = false)
 {
   if(model1.getModelType() != BVH_MODEL_TRIANGLES)
-    return false;
+    throw std::invalid_argument("model1 should be of type BVHModelType::BVH_MODEL_TRIANGLES.");
```
jmirabel, line 311: I understand why one would want this but this makes the API of the `initialize` functions not homogeneous. Either change all of them of none.

I think very few of them return false and you addressed almost all of them already (at least l. 587 is missing), so it shouldn't be a hard job.

jcarpent: Fine for me. I will handle it.

jcarpent: Done.

PR #208, opened 2021-01-20
Fix constness + reduce Boost dependency.

jmirabel:


include/hpp/fcl/profile.h
```diff
@@ -170,6 +169,9 @@ class HPP_FCL_DLLAPI Profiler : private boost::noncopyable
       start();
   }
 
+  Profiler(const Profiler&) = delete;
```
jcarpent, line 172: This is a C++11 feature!

jcarpent (changes requested):
You are forcing the compilation against C++11, which is not something appealing at the current stage of the project, to just remove a minor dependency to Boost.

This is a bit too early I would say.

include/hpp/fcl/profile.h
jmirabel: You are correct. This was not my first intention. This file is anyway not used by anyone. I removed it.

jmirabel:
@jcarpent any idea why windows CI fails ?

jcarpent:
You should have a look to https://github.com/stack-of-tasks/pinocchio/commit/c0a3381de8681c4f38bde1a3dcd7894f300744e1#diff-290e375cf6344cd41694325176f37fab181cde59e19180067c1d37902af69b2c

jmirabel:
@jcarpent if it's okay, can you modify your review ? Now the code is C++98 compliant.

jcarpent (approved):

PR #210, opened 2021-03-08
Add serialization 

jcarpent:


jmirabel:
I am not sure to follow what you did. https://github.com/humanoid-path-planner/hpp-fcl/pull/209 was merged. So it seems all the commits but the last one have been merged already.
Did you do a false manipulation or is it intended ?

jcarpent:
This is a false manip. Sorry for that. I've reverted it.

jcarpent:
By the way, this is still a WIP.

jmirabel:
It is fine if you disable Python doc for Python 2.

jmirabel:
Actually, I think we could solve the problem by running the script always with Python 3. It has nothing to do with the version of the hpp-fcl Python module.

jcarpent:
> Actually, I think we could solve the problem by running the script always with Python 3. It has nothing to do with the version of the hpp-fcl Python module.

Thanks for the tips. I've tried a fix, seems to work on my system. Finger crossed ;)

include/hpp/fcl/BV/AABB.h
```diff
@@ -234,9 +240,9 @@ static inline AABB rotate(const AABB& aabb, const Matrix3f& R)
 {
   AABB res (R * aabb.min_);
   Vec3f corner (aabb.min_);
-  const std::size_t bit[3] = { 1, 2, 4 };
-  for (std::size_t ic = 1; ic < 8; ++ic) { // ic = 0 corresponds to aabb.min_. Skip it.
-    for (std::size_t i = 0; i < 3; ++i) {
+  const Eigen::DenseIndex bit[3] = { 1, 2, 4 };
+  for (Eigen::DenseIndex ic = 1; ic < 8; ++ic) { // ic = 0 corresponds to aabb.min_. Skip it.
+    for (Eigen::DenseIndex i = 0; i < 3; ++i) {
       corner[i] = (ic && bit[i]) ? aabb.max_[i] : aabb.min_[i];
```
jmirabel, line 246: Isn't this bugged ? I would expect this to be
```suggestion
      corner[i] = (ic & bit[i]) ? aabb.max_[i] : aabb.min_[i];
```

include/hpp/fcl/collision_data.h
```diff
@@ -403,7 +411,7 @@ struct HPP_FCL_DLLAPI DistanceResult : QueryResult
                  (std::numeric_limits<FCL_REAL>::max)()):
   min_distance(min_distance_), o1(NULL), o2(NULL), b1(NONE), b2(NONE)
   {
-    Vec3f nan (Vec3f::Constant(std::numeric_limits<FCL_REAL>::quiet_NaN()));
+    static const Vec3f nan (Vec3f::Constant(std::numeric_limits<FCL_REAL>::quiet_NaN()));
```
jmirabel, line 414: Keyword `static` is probably counter-productive. It increases the binary size and I think it can only decrease performances (due to the fact that the variable `nan` cannot be optimized out).

`const` do make sense although I think it doesn't change the compilation result.

include/hpp/fcl/collision_data.h
```diff
@@ -458,23 +466,26 @@ struct HPP_FCL_DLLAPI DistanceResult : QueryResult
   /// @brief clear the result
   void clear()
   {
+    static const Vec3f nan (Vec3f::Constant(std::numeric_limits<FCL_REAL>::quiet_NaN()));
```
jmirabel, line 469: Same here for `static`.

include/hpp/fcl/serialization/eigen.h
```diff
@@ -0,0 +1,74 @@
...
+{
+  namespace serialization
+  {
+    
+    template <class Archive, typename _Scalar, int _Rows, int _Cols, int _Options, int _MaxRows, int _MaxCols>
+    void save(Archive & ar, const Eigen::Matrix<_Scalar,_Rows,_Cols,_Options,_MaxRows,_MaxCols> & m, const unsigned int /*version*/)
+    {
+      Eigen::DenseIndex rows(m.rows()), cols(m.cols());
+      ar & BOOST_SERIALIZATION_NVP(rows);
+      ar & BOOST_SERIALIZATION_NVP(cols);
```
jmirabel, line 49: For compactness reasons, I suggest not to serialize the size when it is known at compile time.

```suggestion
      if (_Rows == Eigen::Dynamic) ar & BOOST_SERIALIZATION_NVP(rows);
      if (_Cols == Eigen::Dynamic) ar & BOOST_SERIALIZATION_NVP(cols);
```

This implies that serialization of Eigen objects in Pinocchio must be updated accordingly.

python/CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ SET(ENABLE_PYTHON_DOXYGEN_AUTODOC TRUE CACHE BOOL "Enable automatic documentatio
 
 IF(    NOT ENABLE_PYTHON_DOXYGEN_AUTODOC
     OR NOT DOXYGEN_FOUND
-    OR (APPLE AND PYTHON_VERSION_MAJOR LESS 3)) # Disable documentation on APPLE for python 2.
+    OR (APPLE AND PYTHON_VERSION_MAJOR LESS 2)) # Disable documentation on APPLE for python 2.
```
jmirabel, line 49: It used to fail with Python 2 on Apple (CI on travis).
If you checked, you should also update or remove the comment. If you didn't, you can let this as is.

jmirabel (commented):
I couldn't see where in the code you actually need the equality operator. Is it only for the unit test or do you have another reason ?

include/hpp/fcl/collision_data.h
jcarpent: The overall overload is very limited in terms of size, it is just a scalar.
Adding constnest could only be positive at this stage.

python/CMakeLists.txt
jcarpent: It is working on my computer, it should not fail as far as I know.
I will remove it now.

include/hpp/fcl/BV/kDOP.h
```diff
@@ -103,6 +103,18 @@ class HPP_FCL_DLLAPI KDOP
   /// @brief Creating kDOP containing two points
   KDOP(const Vec3f& a, const Vec3f& b);
   
+  /// @brief Equality operator
+  bool operator==(const KDOP & other) const
+  {
+    return dist_.isApprox(other.dist_);
```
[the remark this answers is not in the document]
jcarpent: This is not working on Eigen::Array, I don't know why.

include/hpp/fcl/BV/BV_node.h
```diff
@@ -68,6 +68,28 @@ struct HPP_FCL_DLLAPI BVNodeBase
...
+  /// @brief Equality operator
+  bool operator==(const BVNodeBase & other) const
+  {
+    return
+       first_child == other.first_child
+    && first_primitive == other.first_primitive
+    && num_primitives == other.num_primitives;
+  }
+  
+  /// @brief Inequality operator
```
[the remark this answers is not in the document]
jcarpent: Yes, you're totally right.

include/hpp/fcl/BV/AABB.h
jcarpent: I will fix it, even if it is not the scope of this PR initially.

jcarpent:
> I couldn't see where in the code you actually need the equality operator. Is it only for the unit test or do you have another reason ?

Yes, for the tests. And it will be needed also in future inside Pinocchio.

include/hpp/fcl/BV/kDOP.h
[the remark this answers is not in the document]
jmirabel: For arrays, you must do either `(a == b).all()` or `(a == b).any()` depending on your needs.

include/hpp/fcl/BV/AABB.h
jmirabel: Thanks.

jcarpent:
I'm done with this PR. I will add more features later I think.

include/hpp/fcl/BVH/BVH_model.h
```diff
@@ -291,6 +296,74 @@ class HPP_FCL_DLLAPI BVHModel : public BVHModelBase
...
+//    }
+//
+//    if(num_primitives != other_num_primitives)
+//      return false;
+//
+//    for(int k = 0; k < num_primitives; ++k)
+//    {
+//      if(primitive_indices[k] != other.primitive_indices[k])
+//        return false;
+//    }
```
jmirabel, line 348: Any reason to keep this ?

include/hpp/fcl/serialization/BV_fitter.h
```diff
@@ -0,0 +1,20 @@
...
+
+namespace boost
+{
+  namespace serialization
+  {
+  
+  }
+}
+
+#endif // ifndef HPP_FCL_SERIALIZATION_BV_FITTER_H
```
jmirabel, line 20: I think this file can be removed. I don't think we need to serialize the BVH fitter (and BVH splitter).

include/hpp/fcl/BVH/BVH_model.h
jcarpent: Would prefer to keep it for the moment. Because I'm not totally sure if we need to serialize it or not.

include/hpp/fcl/serialization/BV_fitter.h
jcarpent: Yes, you're rigth.

include/hpp/fcl/serialization/eigen.h
```diff
@@ -0,0 +1,89 @@
...
+#include <boost/serialization/vector.hpp>
+#include <boost/serialization/array.hpp>
+
+
+namespace boost
+{
+  namespace serialization
+  {
+    
+#ifndef HPP_FCL_SKIP_EIGEN_BOOST_SERIALIZATION
```
jmirabel, line 25: This preprocessor variable is rather *dangerous* as it can break ABI compatibility.
I guess that the issue was that Pinocchio needs Eigen serialization even without hpp-fcl and that you plan to make Pinocchio defines this variable. This means that Pinocchio has to deal with another preprocessor variable.

Another option, assuming we keep the file identical, is to rely on the header guard (i.e. have the same header guard in Pinocchio and hpp-fcl), for the same file. Optionally, the file could be moved to a git repo added as submodule (the cmake submodule or a dedicated one).

jmirabel: After a private discussion, we will leave this as is.

jmirabel (approved):
I let you merge when ready.

PR #213, opened 2021-03-28
Enable early stopping for GJK solver when computing contacts

jcarpent:
With minor cosmetic change

### DIFF
--- a/include/hpp/fcl/collision_data.h
+++ b/include/hpp/fcl/collision_data.h
@@ -192,7 +192,6 @@ enum CollisionRequestFlag
 {
   CONTACT               = 0x00001,
   DISTANCE_LOWER_BOUND  = 0x00002,
-  EARLY_STOPPING        = 0x00004,
   NO_REQUEST            = 0x01000
 };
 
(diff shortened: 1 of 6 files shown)


include/hpp/fcl/collision_data.h
```diff
@@ -206,6 +207,9 @@ struct HPP_FCL_DLLAPI CollisionRequest : QueryRequest
 
   /// Whether a lower bound on distance is returned when objects are disjoint
   bool enable_distance_lower_bound;
+  
+  /// Whether GJK stops when it proves that the distance between two geometries is greater than distance_upper_bound.
+  bool enable_early_stopping;
```
jmirabel, line 212: I think it is sufficient to set the `distance_upper_bound` to infinity (or max, as in the constructor).

jcarpent: OK. Having the double mechanism seems to be safer at some point. What do you think?

jmirabel (commented):
There are plenty of syntax changes. At the moment, I am not aware of any syntax rules so I would prefer not to include them. You changed some stuff to the way you like and there is nothing that prevents someone to change it later to the way he/she likes.

If we want to do it, we can do it:
- in a separate PR,
- done on all the code,
- by some syntax checker like clang-format.

jmirabel:
And thanks for this change.

include/hpp/fcl/collision_data.h
jmirabel: Having a double mechanism is double-folded. You also need to make sure you updated both.

You can add functions to deal with `distance_upper_bound` if you want this to be *safe* but I don't think it is necessary.

florent-lamiraux:
There are still many modifications of indentation or line breaking that pollute the history. Can you please remove them ?

jcarpent:
Done

florent-lamiraux:
Almost.

PR #219, opened 2021-04-16
Assimp - remove useless LINES and POINTS when loading

jcarpent:


src/mesh_loader/assimp.cpp
```diff
@@ -159,15 +160,7 @@ unsigned recurseBuildMesh (
...
-      if (face.mNumIndices != 3) {
-        std::stringstream ss;
-        ss << "Mesh " << input_mesh->mName.C_Str() << " has a face with "
-           << face.mNumIndices << " vertices. This is not supported\n";
-        ss << "Node name is: " << node->mName.C_Str() << "\n";
-        ss << "Mesh index: " << i << "\n";
-        ss << "Face index: " << j << "\n";
-        throw std::invalid_argument (ss.str());
-      }
+      assert(face.mNumIndices == 3 && "The size of the face is not valid.");
```
wxmerkt, line 163: This will remove this check from compiled versions and may lead to undefined behaviour? The previous code would check and throw an argument. As a non-throwing behaviour, perhaps a warning to std::cerr and continue could be an option if the throwing behaviour is undesired?

jcarpent: From now on, they should not be any more throwing because I change the way assimp is filtering lines and points. I've transformed the previous thrown into an assert for debug purposes.

wxmerkt: Thanks for the clarification. Is it worth adding a test case with a malformed mesh?

jcarpent: Yes, I will.

jcarpent: Done.

jcarpent:
I will make a new release after this merge.

PR #221, opened 2021-04-22
Remove register keyword to fix #220

### LINKED ISSUE #220: C++17 vs register keyword
hpp-fcl currently does not build in workspaces that are set to use `CMAKE_CXX_STANDARD=17` since the `register` keyword is no longer supported. It is e.g. used here:
https://github.com/humanoid-path-planner/hpp-fcl/blob/40c9869a9f9a899afd1512ed4c1fca7cef322a42/src/BV/OBB.cpp#L166

To fix it, we could either (a) fix the code or (b) check that if a standard >14 is requested raise a CMake warning and force a maximum of C++14 for the project. 

What are your thoughts on removing the `register` keyword? Would you consider a PR for that?

wxmerkt:
The register keyword leads to a compile error for C++ standards newer
than or equal to C++17.

### DIFF
--- a/src/BV/OBB.cpp
+++ b/src/BV/OBB.cpp
@@ -163,7 +163,7 @@ inline OBB merge_smalldist(const OBB& b1, const OBB& b2)
 
 bool obbDisjoint(const Matrix3f& B, const Vec3f& T, const Vec3f& a, const Vec3f& b)
 {
-  register FCL_REAL t, s;
+  FCL_REAL t, s;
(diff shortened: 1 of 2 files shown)


jcarpent (approved):
Thanks a lot.

jcarpent:
@wxmerkt Do you need a release?

wxmerkt:
No, I am using source right now - thanks for checking :-).

Issue #227, opened 2021-06-08
Invalid distance_lower_bound


jcarpent:
For simple geometries, like Box and Sphere collision pairs, the distance_lower_bound is not updated even when there is a contact.

jmirabel:
For primitive shapes, the distance lower bound should be the real distance. I'm surprised it is not updated.

When this happens, can you check the distance request works correctly ? Maybe the issue comes from here.

jcarpent:
Solved by #228.

PR #228, opened 2021-06-09
Fix update of distanceLowerBound

jcarpent:
Related to #227


jmirabel (approved):
Indeed, the error was obvious. Thanks !

Issue #201, opened 2020-11-03
Package not distributed with qhull on robotpkg


duburcqa:
Hi,

I just noticed that `hpp-fcl` is not distributed  with `qhull` on `robotpkg` apt-repository, at least for the latest release available. Is it intended ? Personnally, I rely on `qhull` for computing convex hulls and it would be helpful to get it directly without building `hpp-fcl` from source.

Best regards,
Alexis Duburcq

nim65s:
Hi @duburcqa,

I already said I would do it, but this was burried in #167. Thanks for the heads-up, I'll work on that soon.

duburcqa:
Arf I haven't checked the closed issues.  Sorry for the noise. I happy to hear it is on the todo-list !

jmirabel:
@nim65s has this been done ?

nim65s:
not yet, sorry for the delay

nim65s:
fixed in https://github.com/nim65s/robotpkg/commit/0ce391c00588c4f906083a13ff3038e95f691920

PR #233, opened 2021-06-17
[CMake] use system Qhull if available

nim65s:
Hi,

As discussed in #201, I'm activating Qhull on robotpkg. 

Doing so, I noticed that some distributions are already providing qhullcpp, so in that case there is no need to build it. This PR implement this case.

I also checked in other cases, where Qhull is installed but without qhullcpp, that the configuration provided by @jmirabel works well.


jmirabel:
Thanks for handling this !

PR #236, opened 2021-07-30
Extend Python bindings

jcarpent:


jmirabel (approved):
Thanks

jcarpent:
@nim65s Could you check the error on buster on the gitlab-ci? It is not related to this PR, but seems to come from either docker or Boost.

nim65s:
Boost fails for no reason after a successful test on buster, as in many other project. I've allowed failure on this plafform for now.

PR #239, opened 2021-09-05
Add support of stubs generation

jcarpent:


jmirabel (approved):
Assuming you will fix CI, LGTM.

PR #243, opened 2021-09-15
Fix exposition of vertices

jcarpent:


python/collision-geometries.cc
```diff
@@ -474,8 +474,11 @@ void exposeCollisionGeometries ()
     .def ("vertice", &BVHModelBaseWrapper::vertice,
           bp::args("self","index"),"Retrieve the vertex given by its index.",
           bp::return_internal_reference<>())
+    .def ("vertices", &BVHModelBaseWrapper::vertice,
+          bp::args("self","index"),"Retrieve the vertex given by its index.",
+          bp::return_internal_reference<>())
```
jmirabel, line 479: Did you intend to add this ?
It duplicates the function above ("vertice") with the name of the function below ("vertices"). By the way, the singular of *vertices* is *vertex*, not *vertice*.

jcarpent: The thing is that by mistake, I've removed an existing signature used for instance in Pinocchio. 
I need to reintroduce it and I will fix the vertex naming. 
Thanks for the feedback.

jmirabel: Ok. I think it's me who did that. From the user perspective, this looked neat:
```python
obj.vertices(n)
# Looks like
# obj.vertices[n]
```
Now that `obj.vertices[n]` works, we may think of deprecating the first one. This is not a requirement for me. I am fine with keeping both.

jcarpent: I do agree we should deprecate it. We can take inspiration from Pinocchio to do so.

jmirabel (approved):

PR #244, opened 2021-09-16
Enhance Convex exposition

jcarpent:


jcarpent:
Following #243 and #242.

PR #255, opened 2022-01-26
Preparing for ROS1 and ROS2 release

wxmerkt:
WIP - modifications to prepare for a release. Also addressing some warnings along the way and turning others off

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,18 @@ endif()
 
 FIND_PACKAGE(assimp REQUIRED)
 
+# TODO: These warnings should be fixed
+# Ignore sign conversion warnings
+add_compile_options(-Wno-sign-conversion)
+# Ignore deprecated copy
+add_compile_options(-Wno-deprecated-copy)
+# Ignore conversion
+add_compile_options(-Wno-conversion)
+# Ignore sign compare
+add_compile_options(-Wno-sign-compare)
+# Ignore memaccess until replaced with std::copy
+add_compile_options(-Wno-class-memaccess)
+
 SET(${PROJECT_NAME}_HEADERS
   include/hpp/fcl/collision_data.h
   include/hpp/fcl/BV/kIOS.h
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -148,6 +154,18 @@ endif()
 
 FIND_PACKAGE(assimp REQUIRED)
 
+# TODO: These warnings should be fixed
```
jcarpent, line 157: Why do you want to remove that?
I think we need to tackle all the current warnings instead ...

include/hpp/fcl/BVH/BVH_model.h
```diff
@@ -309,23 +309,23 @@ class HPP_FCL_DLLAPI BVHModel
     bool res = Base::operator==(other);
     if(!res)
       return false;
-    
-    unsigned int other_num_primitives = 0;
-    if(other.primitive_indices)
-    {
+
+    // unsigned int other_num_primitives = 0;
```
jcarpent, line 313: Why did you comment here?

jcarpent (changes requested):
Hi @wxmerkt !
I made some comments, could you handle them?
Best,
Justin

CMakeLists.txt
wxmerkt: They should, but there are hundreds. I turned these off to try to get the autodoc-ones resolved (Latex commands seen as escape characters; by changing the Python), but wasn't successful initially so stopped.

include/hpp/fcl/BVH/BVH_model.h
wxmerkt: It's set and not used -- the code below where it would be used is already commented

CMakeLists.txt
jcarpent: @wxmerkt Could you remove these lines?

include/hpp/fcl/BVH/BVH_model.h
jcarpent: Could you revert this change?

wxmerkt: If you take a look at the wider context of this function, the result of these lines is never used -- the lines that originally used them are commented below. Is there a reason to leave it in there if it's unused?

jmirabel: I agree it should have been commented in the first place.

jcarpent (approved):

PR #256, opened 2022-01-27
[OcTree] Add method to save octree in obj file.

florent-lamiraux:


jmirabel (approved):
A unit test would be nice if you have time.

src/octree.cpp
```diff
@@ -38,6 +38,150 @@ namespace hpp
...
+} // namespace internal
+
+  void OcTree::exportAsObjFile(const std::string& filename) const
+  {
+    std::vector<boost::array<FCL_REAL, 6> > boxes(this->toBoxes());
+    std::vector<internal::Neighbors> neighbors(boxes.size());
+    internal::computeNeighbors(boxes, neighbors);
+    // compute list of vertices and faces
+    std::vector<std::array<FCL_REAL, 3> > vertices;
+    std::vector<std::array<std::size_t, 4> > faces;
```
jmirabel, line 136: This makes the code C++11. I think there are still people requiring C++98 (not sure what for though).

src/octree.cpp
```diff
@@ -38,6 +38,150 @@ namespace hpp
...
+    }
+    // write obj in a file
+    std::ofstream os;
+    os.open(filename);
+    if (!os.is_open())
+      throw std::runtime_error(std::string("failed to open file \"")+
+			       filename + std::string("\""));
+      // write vertices
+    os << "# list of vertices" << std::endl;
+    for (auto v : vertices){
```
jmirabel, line 175: Same here.

PR #258, opened 2022-02-08
Fix C++98 compatibility

jcarpent:


src/octree.cpp
```diff
@@ -40,165 +40,200 @@ namespace fcl
...
+    const Vec3f & v = *it;
+    os << "v " << v[0] << " " << v[1] << " " << v[2] << std::endl;
+  }
+  os << std::endl << "# list of faces" << std::endl;
+  for (VectorArray4::const_iterator it = faces.begin();
+       it != faces.end(); ++it)
+  {
+    const Array4 & f = *it;
+    os << "f " << f[0] << " " << f[1] << " " << f[2] << " " << f[3] << std::endl;
+  }
```
jmirabel, line 218: I know this is not the scope of this PR but, for large octrees, the `std::endl` will be very inefficient.

```suggestion
  os << "# list of vertices\n";
  for (VectorVec3f::const_iterator it = vertices.begin();
       it != vertices.end(); ++it)
  {
    const Vec3f & v = *it;
    os << "v " << v[0] << " " << v[1] << " " << v[2] << '\n';
  }
  os << "\n# list of faces\n";
  for (VectorArray4::const_iterator it = faces.begin();
       it != faces.end(); ++it)
  {
    const Array4 & f = *it;
    os << "f " << f[0] << " " << f[1] << " " << f[2] << " " << f[3] << '\n';
  }
```

jmirabel (approved):
Never ending discussion: I don't think it makes much sense to re-indent the code given
- given that we do not have any guidelines and
- without enforcing it in the first place (i.e. at PR time).

jcarpent:
After the CI is green, I will make a new release.

jcarpent:
After that, we can switch to C++11.
I will also reintroduce the desperate broadphase ;)

PR #259, opened 2022-02-14
Add Ellipsoid

jcarpent:
This PR done by @lmontaut introduces the class Ellipsoid, which can be further used for external applications.

### DIFF
--- a/src/narrowphase/gjk.cpp
+++ b/src/narrowphase/gjk.cpp
@@ -129,6 +129,19 @@ inline void getShapeSupport(const Sphere*, const Vec3f& /*dir*/, Vec3f& support,
   support.setZero();
 }
 
+inline void getShapeSupport(const Ellipsoid* ellipsoid, const Vec3f& dir, Vec3f& support, int&, MinkowskiDiff::ShapeData*)
+{
+  FCL_REAL a2 = ellipsoid->radii[0] * ellipsoid->radii[0];
+  FCL_REAL b2 = ellipsoid->radii[1] * ellipsoid->radii[1];
+  FCL_REAL c2 = ellipsoid->radii[2] * ellipsoid->radii[2];
+
+  Vec3f v(a2 * dir[0], b2 * dir[1], c2 * dir[2]);
+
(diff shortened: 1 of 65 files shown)


jcarpent:
I will merge it as there is no conflict with the existing work.

src/narrowphase/gjk.cpp
```diff
@@ -381,6 +397,10 @@ MinkowskiDiff::GetSupportFunction makeGetSupportFunction1 (const ShapeBase* s1,
     inflation[1] = static_cast<const Sphere*>(s1)->radius;
     if (identity) return getSupportFuncTpl<Shape0, Sphere, true >;
     else          return getSupportFuncTpl<Shape0, Sphere, false>;
+  case GEOM_ELLIPSOID:
+    // TODO: what should we put for inflation[1]?
```
jmirabel, line 401: It should be 0 since the support function returns a point on the shape. I think it is initialized to 0.

src/narrowphase/gjk.cpp
```diff
@@ -421,6 +441,10 @@ MinkowskiDiff::GetSupportFunction makeGetSupportFunction0 (const ShapeBase* s0,
     inflation[0] = static_cast<const Sphere*>(s0)->radius;
     return makeGetSupportFunction1<Sphere> (s1, identity, inflation, linear_log_convex_threshold);
     break;
+  case GEOM_ELLIPSOID:
+    // TODO: what should we put for inflation[0]?
```
jmirabel, line 445: Same as above.

src/narrowphase/gjk.cpp
jcarpent: @lmontaut Could you handle the comment made by @jmirabel within a new PR?

lmontaut: Yes no problem.

lmontaut: I simply removed the comment, inflation is already set to 0 as mentioned by @jmirabel 
See #261

PR #262, opened 2022-02-16
Address some warnings

wxmerkt:
Uses C++11 features which I expect will be default after #260 merges

Following this PR, the only warnings I see locally are from the Python-autodoc

### DIFF
--- a/src/broadphase/broadphase_dynamic_AABB_tree_array.cpp
+++ b/src/broadphase/broadphase_dynamic_AABB_tree_array.cpp
@@ -507,9 +507,9 @@ bool distanceRecurse(DynamicAABBTreeArrayCollisionManager::DynamicAABBNode* node
 
 //==============================================================================
 DynamicAABBTreeArrayCollisionManager::DynamicAABBTreeArrayCollisionManager()
-  : tree_topdown_balance_threshold(&dtree.bu_threshold),
-    tree_topdown_level(&dtree.topdown_level)
 {
+  tree_topdown_balance_threshold = &dtree.bu_threshold;
+  tree_topdown_level = &dtree.topdown_level;
   max_tree_nonbalanced_level = 10;
   tree_incremental_balance_pass = 10;
   *tree_topdown_balance_threshold = 2;
--- a/src/narrowphase/gjk.cpp
+++ b/src/narrowphase/gjk.cpp
@@ -224,7 +224,7 @@ void getShapeSupportLog(const ConvexBase* convex, const Vec3f& dir, Vec3f& suppo
   FCL_REAL maxdot = pts[hint].dot(dir);
   std::vector<int8_t>& visited = data->visited;
   visited.assign(convex->num_points, false);
-  visited[(size_t)hint] = true;
+  visited[static_cast<std::size_t>(hint)] = true;
   // when the first face is orthogonal to dir, all the dot products will be
   // equal. Yet, the neighbors must be visited.
   bool found = true, loose_check = true;
@@ -245,7 +245,7 @@ void getShapeSupportLog(const ConvexBase* convex, const Vec3f& dir, Vec3f& suppo
         better = true;
       if (better) {
         maxdot = dot;
-        hint = (int)ip;
+        hint = static_cast<int>(ip);
         found = true;
       }
     }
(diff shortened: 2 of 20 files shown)


jcarpent (approved):
Seems good to me.

wxmerkt:
Btw - before merging this, we need to merge https://github.com/jrl-umi3218/jrl-cmakemodules/pull/504 and then update cmake to the updated master

wxmerkt:
The "collision" test fails for me under clang both in master and in this branch. Is this a known issue? It appears fine using gcc

jcarpent:
It is Working well on my side with clang. Could you prepare a docker image for this issue?

wxmerkt:
This PR should be ready for a full review once CI comes back green.

The crash was due to it picking up an ABI incompatible copy of the library in another directory (likely an incorrect LD_LIBRARY_PATH at config time). Interestingly though, the error did not show with gcc. In either case, it's working in a Docker. Strangely enough though, the same error showed up in the Clang-ROS-CI-Config I just added (https://github.com/humanoid-path-planner/hpp-fcl/runs/5224552889?check_suite_focus=true#step:4:1755). It doesn't show if it's not sourced (which was required for Crocoddyl's CI setup); but that shouldn't mean much since it's a different build system anyways.

jcarpent (changes requested):
@wxmerkt Thanks for this PR. Could you remove the reverted commits? They are not informative and seems to pollute the commit history.

src/broadphase/broadphase_dynamic_AABB_tree_array.cpp
```diff
@@ -507,9 +507,9 @@ bool distanceRecurse(DynamicAABBTreeArrayCollisionManager::DynamicAABBNode* node
 
 //==============================================================================
 DynamicAABBTreeArrayCollisionManager::DynamicAABBTreeArrayCollisionManager()
-  : tree_topdown_balance_threshold(&dtree.bu_threshold),
-    tree_topdown_level(&dtree.topdown_level)
 {
+  tree_topdown_balance_threshold = &dtree.bu_threshold;
```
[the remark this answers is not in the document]
wxmerkt: dtree is used uninitialised - and adding a initialisation in the declaration doesn't address the issue since the int-ptr are declared before dtree is declared. By moving this line into the constructor, dtree is correctly initialised before the pointer is assigned

src/narrowphase/gjk.cpp
```diff
@@ -739,7 +739,7 @@ bool GJK::encloseOrigin()
   switch(simplex->rank)
   {
   case 1:
-    for(int i = 0; i < 3; ++i)
+    for(Vec3f::Index i = 0; i < 3; ++i)
```
jcarpent, line 742: int is fine here.

src/narrowphase/gjk.cpp
```diff
@@ -755,7 +755,7 @@ bool GJK::encloseOrigin()
   case 2:
     {
       Vec3f d = simplex->vertex[1]->w - simplex->vertex[0]->w;
-      for(int i = 0; i < 3; ++i)
+      for(Vec3f::Index i = 0; i < 3; ++i)
```
jcarpent, line 758: Same here.

src/narrowphase/gjk.cpp
jcarpent: `Index` is not supported by all Eigen version. One should rather rely on `DenseIndex`.

wxmerkt: Indeed. I believe it was size_t when I started the branch and this is from resolving a rebase conflict. Should I revert? Is this a problem to change it to Vec3f::Index (which should resolve to the same)?

jcarpent: The only thing with Index, it is that it has been introduced lately in Eigen. But I think it was in Eigen 3.3.0 if I remember well.

wxmerkt: Do we have a supported configuration where Eigen is <3.3 (it has been released in Nov 2016)?

jcarpent: I would say yes, if has no cost. I think @nim65s still have some CI with 16.04.

wxmerkt: Ok, addressed

jcarpent (approved):
Thanks a lot @wxmerkt.

jcarpent:
After the CI is green, I will merge this PR.

PR #264, opened 2022-02-21
Fix missing copy of aabb_local in CollisionGeometry

jcarpent:


include/hpp/fcl/collision_object.h
```diff
@@ -79,6 +79,7 @@ class HPP_FCL_DLLAPI CollisionGeometry
   CollisionGeometry(const CollisionGeometry & other)
   : aabb_center(other.aabb_center)
   , aabb_radius(other.aabb_radius)
+  , aabb_local(other.aabb_local)
```
jmirabel, line 82: Do we have a good reason not to have `CollisionGeometry(const CollisionGeometry & other) = default` ?

jcarpent: No, except before it was C++98 default ;)

jmirabel: Yes, but it could just not have been there.
Thanks for the update.

jmirabel (approved):

PR #265, opened 2022-02-22
broadphase: set all methods virtual for inheritence

jcarpent:


jcarpent:
Maybe this is not a good idea. I will see if I can do something else.

PR #266, opened 2022-03-01
use std::shared_ptr, fix #218

### LINKED ISSUE #218: Moving to `std::shared_ptr`
Hi,

The HPP framework is now using C++11, and is therefore going move to `std::shared_ptr` instead of `boost`ones.

Keeping `boost::shared_ptr` in hpp-fcl would be a pain, and this might not be totally necessary.

I suggest to move hpp-fcl to C++11 and `std::shared_ptr`, and to release a new major version for this change.

For pinocchio, and other third-parties that rely on hpp-fcl, this might lead to something like:
https://github.com/stack-of-tasks/pinocchio/blob/master/src/CMakeLists.txt#L51-L54

Obviously, I can handle this work, and the related updates in the HPP, SoT, loco-3d & gepetto projects, but **is there any objection for this** ?

My main motivation is that the current version of hpp-pinocchio is not compatible with pinocchio v2.6.0, so I can't publish pinocchio v2.6.0 on robotpkg until a new HPP release is done, and the switch to std::shared_ptr is already on-going on this new HPP release.

nim65s:


jcarpent (approved):
Should we enforce the use of the `std` namespace?

nim65s:
> Should we enforce the use of the `std` namespace?

I am not sure about what you mean.
If you mean "using all the `std` namespace", I think this is not a good idea
If you mean "forbidding `boost::shared_ptr`", I think this is a good idea, but I don't see how we could do that

PR #268, opened 2022-03-02
Fix broadphase warnings for clang (some conversion remain for g++)

wxmerkt:


jcarpent (commented):
I don't have any warning on my side. Could you give me the compiler and its version in use?

wxmerkt:
clang10 on Ubuntu 20.04:
```
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
```

jcarpent:
OK. I will accept it, even if I think this is a mistake from your compiler or maybe the Eigen version in use.
Normally, such an operation should not need a cast as the vector is made of double values.

jcarpent (approved):

jmirabel:
By curiosity, can you check if you have a warning with an int instead of a long.

jmirabel:
Then I think it depends on the size of long your compiler uses.

Although minor to me and questionable, I think it makes more sense to use int (with a static cast). I don't think we will ever reach 2^31 leaves (except may be on very particular setup).

jcarpent:
I do think that the issue is coming from your compiler. I don't have such warnings with clang 13.0

jmirabel:
It does not depend only on the compiler. If you want to be sure it's the compiler faults, you should check that `sizeof(long)` is the same with both setup.

PR #271, opened 2022-03-10
[ComputeCollision] Return no collision if security_margin is set to -inf

florent-lamiraux:
This implements https://github.com/humanoid-path-planner/hpp-fcl/issues/269.

### DIFF
--- a/src/collision.cpp
+++ b/src/collision.cpp
@@ -78,6 +78,11 @@ std::size_t collide(const CollisionGeometry* o1, const Transform3f& tf1,
                     const CollisionGeometry* o2, const Transform3f& tf2,
                     const CollisionRequest& request, CollisionResult& result)
 {
+  // If securit margin is set to -infinity, return that there is no collision
+  if (request.security_margin == -std::numeric_limits<FCL_REAL>::infinity()) {
+    result.clear();
+    return false;
+  }
   GJKSolver solver;
   solver.enable_cached_guess = request.enable_cached_gjk_guess;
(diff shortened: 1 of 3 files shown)


src/collision.cpp
```diff
@@ -163,6 +163,11 @@ ComputeCollision::ComputeCollision(const CollisionGeometry* o1,
 std::size_t ComputeCollision::run(const Transform3f& tf1, const Transform3f& tf2,
        const CollisionRequest& request, CollisionResult& result) const
 {
+  // If securit margin is set to -infinity, return that there is no collision
+  if (request.security_margin == -std::numeric_limits<FCL_REAL>::infinity()) {
+    result.clear();
+    return false;
+  }
```
jmirabel, line 170: Shouldn't you also modify function `collision` ?

florent-lamiraux: You probably mean function `collide`. You are right. I have modified it.

jmirabel (approved):
Could you also add a note in the doc somewhere ?
Not sure where it best fit. Maybe here https://github.com/humanoid-path-planner/hpp-fcl/blob/47941ac4cdb8c72c2640babd73f39812946ae302/include/hpp/fcl/collision_data.h#L222
or in this section https://github.com/humanoid-path-planner/hpp-fcl/blob/47941ac4cdb8c72c2640babd73f39812946ae302/include/hpp/fcl/doc.hh#L58

jcarpent (approved):

PR #274, opened 2022-03-16
Add operator== and operator!= to CollisionGeometry

jcarpent:


include/hpp/fcl/collision_object.h
```diff
@@ -93,14 +94,14 @@ class HPP_FCL_DLLAPI CollisionGeometry
     && aabb_center == other.aabb_center
     && aabb_radius == other.aabb_radius
     && aabb_local == other.aabb_local
-//    && user_data == other.user_data
+    && typeid(*this) == typeid(other) && isEqual(other)
```
jmirabel, line 97: Using `typeid` for this is a bit dangerous. `typeid` is not consistent across different binary object. I am not 100% sure this will not cause subtle bug in dependencies.
Are you aware of this ? Any thoughts ?
The safest, but also more verbose, if dynamic casting in `isEqual` implementations.

jcarpent: Yes, I was aware of this dangerously but I was thinking it was not dangerous in the current case
I can switch to dynamic_cast indeed ...

jcarpent: I know that in Boost.Python they provide a workaround for that ...
here is an interesting discussion https://www.axelmenzel.de/articles/rtti

jmirabel: > In some special situation the typeid operator doesn't work, more precisely there are problems when used [across shared boundaries.](https://svn.boost.org/trac/boost/ticket/754) In C++ the concept of shared or dynamic libraries is not defined, so it is not a problem with the standard. Nevertheless, there exist a workaround for this problem:
**Using the type_info::name function for comparison instead of the built in compare operator will fix this issue**, but this is of course slower. Another drawback is that the type_info objects are not copyable. That means it is not possible to use them directly in a container. Creating a [wrapper class](http://loki-lib.sourceforge.net/html/a00657.html) around type_info object will fix this issue.

Thanks for the link. Above is a relevant paragraph from your link. I didn't know the workaround (in bold above).

jcarpent: I applied the change in ae893c6

include/hpp/fcl/collision_object.h
```diff
@@ -93,14 +94,14 @@ class HPP_FCL_DLLAPI CollisionGeometry
...
-//    && user_data == other.user_data
+    && isEqual(other)
     ;
   }
 
   /// \brief Difference operator
   bool operator!=(const CollisionGeometry & other) const
   {
-    return !(*this == other);
+    return typeid(*this) == typeid(other) && isNotEqual(other);
```
jmirabel, line 104: I guess now it should be
```suggestion
    return isNotEqual(other);
```

include/hpp/fcl/collision_object.h
jmirabel: Thanks.

include/hpp/fcl/collision_object.h
jcarpent: Thanks

PR #275, opened 2022-03-20
Fix some missing features in base classes

jcarpent:


include/hpp/fcl/shape/details/convex.hxx
```diff
@@ -98,11 +98,15 @@ template <typename PolygonT>
 Convex<PolygonT> * Convex<PolygonT>::clone() const
 {
   Vec3f * cloned_points = new Vec3f[num_points];
-  memcpy(cloned_points, polygons, sizeof(Vec3f) * num_points);
+  memcpy(cloned_points, points, sizeof(Vec3f) * num_points);
```
wxmerkt, line 101: Using `memcpy` will introduce a warning - can we use `std::copy`?

jcarpent: Yes, sure. Thanks for the report.

jmirabel (approved):

PR #278, opened 2022-03-24
Add operator{==,!=} to CollisionObject

jcarpent:


include/hpp/fcl/collision.h
```diff
@@ -112,6 +112,20 @@ class HPP_FCL_DLLAPI ComputeCollision
     return res;
   }
   
+  bool operator==(const ComputeCollision & other) const
+  {
+    return o1 == other.o1
+    && o2 == other.o2
+    && swap_geoms == other.swap_geoms
+    && solver == other.solver
+    && func == other.func;
```
jmirabel, line 121: Side remark:
In theory, this should be sufficient.
```suggestion
    return o1 == other.o1
    && o2 == other.o2
    && solver == other.solver;
```
The two other attributes are just caching redundant info.

Maybe it's better to test them anyway.

jmirabel (approved):

Issue #279, opened 2022-04-02
Formatting commits via pre-commit


jcarpent:
@nim65s, @florent-lamiraux, @jmirabel 

With @nim65s, we have been discussing using pre-commit to force the formatting of the projects.
This will enable to have the same standard to apply to everybody.

I will open a PR in this direction, and the rules should be effective for hpp-fcl 2.0
Is it fine for everybody?

jmirabel:
You mean applying clang format before committing. Does it mean you have to have clang format to be able to commit ?
Do you plan to apply code formatting to the whole repo ?

jcarpent:
> Does it mean you have to have clang format to be able to commit ?

You will be able to configure the commit formatting via `pre-commit install`.
A bot will check if the commits comply with the internal project rules.

> Do you plan to apply code formatting to the whole repo ?

I would say yes. With a single commit line.

nim65s:
> You mean applying clang format before committing. 

For people who run `pre-commit install`, yes.

> Does it mean you have to have clang format to be able to commit ?

No. If you use `pre-commit install`, pre-commit will install clang-format in a venv made for pre-commit.

And if somebody doesn't have `pre-commit`, or doesn't run `pre-commit install`, it's not an issue: pre-commit CI will add formatting commits in every PR from that person.

> Do you plan to apply code formatting to the whole repo ?

It would be better

florent-lamiraux:
If it is automatic, I agree.

jcarpent:
Solved with #280.

Issue #267, opened 2022-03-02
Hidden virtual in Python bindings overload for `distance`


wxmerkt:
This warning appears to indicate an issue: The current Python API omits the third (out) parameter, the distance. Hence, the overload of `distance` doesn't actually return the distance:
```
In file included from /git/hpp-fcl/python/broadphase/broadphase.cc:58:
/git/hpp-fcl/python/broadphase/broadphase_callbacks.hh:82:8: warning: 'hpp::fcl::DistanceCallBackBaseWrapper::distance' hides overloaded virtual function [-Woverloaded-virtual]
  bool distance(CollisionObject* o1, CollisionObject* o2) { return this->get_override("distance")(o1,o2); }
       ^
/git/hpp-fcl/include/hpp/fcl/broadphase/broadphase_callbacks.h:80:16: note: hidden overloaded virtual function 'hpp::fcl::DistanceCallBackBase::distance' declared here: different number of parameters (3 vs 2)
  virtual bool distance(CollisionObject* o1, CollisionObject* o2, FCL_REAL & dist) = 0;
               ^
1 warning generated.
```

I might misinterpret the intent here - this doesn't appear to return the distance at all. To fix this, would a change of the Python API to `double distance(o1, o2)` be appropriate? Or should we add the `FCL_REAL& dist` as a parameter to fix it? Any other suggestions?

jcarpent:
This is still a work in progress. Please be patient.

jcarpent:
Thanks @wxmerkt for reminding me of this issue. We should keep the inout argument solution as the distance is updated.
The solution would consist in implementing the two signatures.

PR #280, opened 2022-04-02
Configure and apply pre-commit

jcarpent:
Related to #279.


nim65s:
While we're at it, I think we should also apply black to reformat python code, by adding this section to `.pre-commit-config.yaml`:
```yaml
-   repo: https://github.com/psf/black
    rev: 22.3.0
    hooks:
    -   id: black
```

jmirabel:
I would like that people looking at `git blame` understand that they are looking at a code reformatting commit and not an actual change. To this end, I propose to
- to reword commit *pre-commit: init all* to something more explicit. Maybe *Reformat code to Google style*
- modify the author of the message to *(Reformat) Justin Carpentier*. I am asking because depending on how you configure `git blame`, you don't have the commit message right away.

nim65s:
We can also exclude these reformating commits from git blame

jmirabel:
> We can also exclude these reformating commits from git blame, with `--ignore-rev` / `--ignore-revs-file`. A `.git-blame-ignore-revs` file, which can be versioned in the repo, seems to a common convention: https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame / https://akrabat.com/ignoring-revisions-with-git-blame/

Didn't know about this. Indeed, that makes code reformatting a lot more attractive !

nim65s:
Then I suggest we merge this, and release. Should I handle that ?

jmirabel:
Don't you want to add the git blame file first ?
Then, yes, it's good for me.

nim65s:
sure

nim65s:
PS: Github support for this feature is still in beta: https://github.com/github/feedback/discussions/5033

wxmerkt:
Should we expand the README on how to use pre-commit and to manually set the git-blame setting (per commit message it needs to be set manually on every developer machine)?
```
[add git blame ignore file](https://github.com/humanoid-path-planner/hpp-fcl/pull/280/commits/5fea4840a47ec1b895757f6ecc5bdc691ae1535a)
[5fea484](https://github.com/humanoid-path-planner/hpp-fcl/pull/280/commits/5fea4840a47ec1b895757f6ecc5bdc691ae1535a)

To use it, either:
- git blame --ignore-revs-file .git-blame-ignore-revs
- git config blame.ignoreRevsFile .git-blame-ignore-revs
```

nim65s:
I'm not sure how useful it would be to add more docs about those standard tools here. 

They are becoming pretty common, and for anybody who is not aware of those, github UI will automatically provide the right blame, and pre-commit.ci will automatically add commits to re-format if needed. 

But I won't oppose if someone provide that doc :)

jcarpent:
I can handle the release soon also.

jcarpent:
Before doing the new release, I will fix this issue #267.

jcarpent:
The new release version will thus be 2.0.0

PR #289, opened 2022-05-04
Feature: Nesterov acceleration for GJK

lmontaut:
This PR implements a Nesterov acceleration for the GJK algorithm. 
In `gjk.cpp`, we simply replace the direction used to compute support points from `ray` to `dir`, where `dir` is obtained using the Nesterov scheme. An enum allows to switch from `Default` GJK to `NesterovAcceleration` GJK.

The python bindings are modified accordingly.

In a following PR I will put a link to the corresponding paper.

### DIFF
--- a/include/hpp/fcl/narrowphase/gjk.h
+++ b/include/hpp/fcl/narrowphase/gjk.h
@@ -148,6 +148,8 @@ struct HPP_FCL_DLLAPI GJK {
   };
 
   enum Status { Valid, Inside, Failed };
+  /// @brief Version used for GJK. Vanilla is default.
+  enum GJKVariant { Vanilla, Nesterov };
 
   MinkowskiDiff const* shape;
   Vec3f ray;
@@ -228,6 +230,14 @@ struct HPP_FCL_DLLAPI GJK {
     distance_upper_bound = dup;
   }
 
+  /// @brief Set which GJK version to use. Default is Vanilla.
+  inline void setGJKVariant(GJKVariant variant) { gjk_variant = variant; }
+
+  /// @brief Set wether or not to use the normalization heuristic when computing a support point.
+  /// Only effective if acceleration version of GJK is used.
+  /// Default is false.
+  inline void setNormalizeSupportDirection(bool normalize) { normalize_support_direction = normalize; }
+
  private:
   SimplexV store_v[4];
   SimplexV* free_v[4];
@@ -239,6 +249,8 @@ struct HPP_FCL_DLLAPI GJK {
   unsigned int max_iterations;
   FCL_REAL tolerance;
   FCL_REAL distance_upper_bound;
+  GJKVariant gjk_variant;
+  bool normalize_support_direction;
 
   /// @brief discard one vertex from the simplex
   inline void removeVertex(Simplex& simplex);
--- a/include/hpp/fcl/narrowphase/gjk.h
+++ b/include/hpp/fcl/narrowphase/gjk.h
@@ -149,7 +149,7 @@ struct HPP_FCL_DLLAPI GJK {
 
   enum Status { Valid, Inside, Failed };
   /// @brief Version used for GJK. Vanilla is default.
-  enum GJKVariant { Vanilla, Nesterov };
+  enum GJKVariant { Default, NesterovAcceleration };
 
   MinkowskiDiff const* shape;
   Vec3f ray;
--- a/include/hpp/fcl/narrowphase/gjk.h
+++ b/include/hpp/fcl/narrowphase/gjk.h
@@ -238,6 +238,9 @@ struct HPP_FCL_DLLAPI GJK {
   /// Default is false.
   inline void setNormalizeSupportDirection(bool normalize) { normalize_support_direction = normalize; }
 
+  /// @brief Get GJK number of iterations.
+  inline size_t getIterations() { return iterations; }
+
  private:
   SimplexV store_v[4];
   SimplexV* free_v[4];
@@ -251,6 +254,7 @@ struct HPP_FCL_DLLAPI GJK {
   FCL_REAL distance_upper_bound;
   GJKVariant gjk_variant;
   bool normalize_support_direction;
+  size_t iterations;
 
   /// @brief discard one vertex from the simplex
   inline void removeVertex(Simplex& simplex);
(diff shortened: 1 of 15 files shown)


jcarpent (approved):

nim65s:
nice, thanks :)

A test checking twice the same distance with both variants and comparing the number of iterations would be interesting, don't you think ?

jcarpent:
You are right @nim65s. A nice paper is coming on the topic with a very huge number of tests.
@lmontaut could you have a sanity check to validate the convergence of the variant?

lmontaut:
Yes thanks @nim65s ! I will add tests with different collision pairs to check that the solution found by vanilla GJK and GJK + Nesterov are the same.

Regarding reducing the number of iterations, I will link the paper with all the benchmarks. The number of iteration is reduced when objects are in close-proximity, basically when the broadphase fails to say that the objects are not in collision and thus the narrowphase needs to be called.
If we add a test to check that Nesterov does less iterations than GJK, it will sometimes fail. In the paper, we show that a vast majority of collision problems benefit the Nesterov acceleration, but in some cases Nesterov will actually add a few iterations to GJK.

Anyway, I can also include a test which verifies that Nesterov + GJK converges in a reasonable number of iterations, like GJK.

src/narrowphase/gjk.cpp
```diff
@@ -687,12 +718,17 @@ GJK::Status GJK::evaluate(const MinkowskiDiff& shape_, const Vec3f& guess,
...
-      if (distance < tolerance) status = Inside;
-      break;
+      if (current_gjk_variant != Default) {
+        current_gjk_variant = Default;
+        continue;
+      } else {
+        distance = rl - inflation;
+        // TODO When inflation is strictly positive, the distance may be exactly
+        // zero (so the ray is not zero) and we are not in the case rl <
+        // tolerance.
```
jmirabel, line 728: Is this a problem ?

include/hpp/fcl/narrowphase/gjk.h
```diff
@@ -228,6 +230,19 @@ struct HPP_FCL_DLLAPI GJK {
     distance_upper_bound = dup;
   }
 
+  /// @brief Set which GJK version to use. Default is Vanilla.
+  inline void setGJKVariant(GJKVariant variant) { gjk_variant = variant; }
+
+  /// @brief Set wether or not to use the normalization heuristic when computing
+  /// a support point. Only effective if acceleration version of GJK is used.
+  /// Default is false.
+  inline void setNormalizeSupportDirection(bool normalize) {
```
jmirabel, line 239: Shouldn't the default value depend on the type of object pair ?

jcarpent: At first glance, it should be better to keep the vanilla GJK as the default variant. We should enforce the use of the accelerated variant for proximity requests, like after culling using broadphases.

src/narrowphase/gjk.cpp
```diff
@@ -663,20 +673,54 @@ GJK::Status GJK::evaluate(const MinkowskiDiff& shape_, const Vec3f& guess,
...
+        dir = ray;
+        break;
+
+      case NesterovAcceleration:
+        // Normalize heuristic for collision pairs involving convex but not
+        // strictly-convex shapes This corresponds to most use cases.
+        if (normalize_support_direction) {
+          momentum = (FCL_REAL(iterations) + 2) / (FCL_REAL(iterations) + 3);
+          y = momentum * ray + (1 - momentum) * w;
+          dir = momentum * dir / dir.norm() + (1 - momentum) * y / y.norm();
```
jcarpent, line 688: It would be nice to add an assert to check that the y.norm() is above a dummy_precision threshold.

src/narrowphase/gjk.cpp
jcarpent: @lmontaut Could you reply?

include/hpp/fcl/narrowphase/gjk.h
lmontaut: @jcarpent the default value of `normalize_support_direction` does not affect the default variant used for GJK. The default variant for GJK is the vanilla one, it is set in the `initialize` function called in GJK's constructor

lmontaut: Apologies for the late reply, I answered a few days ago but didn't see that my comment was "Pending" and so was only visible to me...

src/narrowphase/gjk.cpp
jmirabel: I was wondering why this is a TODO. From your answer, it seems this is a comment that does not need a code change.

src/narrowphase/gjk.cpp
lmontaut: Done. I added a comment to explain the assertion check.

src/narrowphase/gjk.cpp
lmontaut: My bad, I remove this code change from the PR!

include/hpp/fcl/narrowphase/gjk.h
lmontaut: I am putting this PR on hold as in the test I am doing and planning to add for the Nesterov acceleration, I am realising that as @jmirabel suggested, the default value for the normalize heuristic does depend on the type of object pair. To sum up, any pair that has a strictly convex object like an ellipsoid should have the heuristic turned off.

I think that a solution could be to add a matrix just like in `collision_func_matrix.h` . For each pair of geometries which has a support function implemented in `gjk.cpp` (triangle, cone, box, cylinder, ellipsoid, sphere, box, mesh), have the heuristic set to true or false by default. Does this sound like a good idea?

jmirabel: I don't think you need a matrix. In `src/narrowphase/gjk.cpp`, you can add an enum to each `shape_traits` to say whether this shape is strictly convex or not.
Then, in `MinkowskiDiff`, by checking whether the Minkowski difference is strictly convex, store the information and used it wherever you need it.

lmontaut: Thanks @jmirabel !

lmontaut:
Ok I think this PR is ready for reviews. Apologies for taking a long time.

I have added tests for the nesterov acceleration. In the test utility, I added a function to construct a 20 faces mesh (`Convex<Triangle>`) to test the solvers on it. I found it useful when I did my tests for the paper, but I can of course remove it.

jcarpent (approved):

PR #290, opened 2022-05-04
Add enum helpers

jcarpent:


nim65s (approved):

include/hpp/fcl/collision_utility.h
```diff
@@ -21,11 +22,38 @@
...
+      "GEOM_HALFSPACE", "GEOM_TRIANGLE", "GEOM_OCTREE", "GEOM_ELLIPSOID",
+      "HF_AABB",        "HF_OBBRSS",     "NODE_COUNT"};
+
+  return node_type_name_all[node_type];
 }
 
+/**
+ * \brief Returns the name associated to a OBJECT_TYPE
+ */
+inline const char* get_object_type_name(OBJECT_TYPE object_type) {
```
jmirabel, line 48: You can return a `std::string` here. In C++17, you may be able to use `constexpr` as well (not 100% sure).

jmirabel (approved):
That makes the error messages more readable. Thanks.

include/hpp/fcl/collision_utility.h
jcarpent: I guess it is fined only to consider const char *. No?

jmirabel: I suggested this just because you wrap it in `std::string` everywhere you use it. I am fine with `const char*` too.

PR #299, opened 2022-05-24
Feature: adding convergence criterions for GJK algorithm

lmontaut:
This PR adds the following:
- 2 more convergence criterions in addition to the default hppfcl criterion, in `GJK`. These provide different guarantees on the distance between the solution found by GJK and the optimal solution (the projection of the origin onto the Minkowski difference). Currently, hppfcl is using the criterion from the 1999 Van den Bergen [paper](http://www.dtecta.com/papers/jgt98convex.pdf). This PR adds the frank-wolf duality gap (used in the original GJK paper) as well as a mix between VDB and the duality-gap.
- exposing them to the collision/distance api, and to python
- tests for these criterions

### DIFF
--- a/test/gjk_convergence_criterion.cpp
+++ b/test/gjk_convergence_criterion.cpp
@@ -49,114 +49,121 @@
 
 using hpp::fcl::Box;
 using hpp::fcl::FCL_REAL;
+using hpp::fcl::GJKConvergenceCriterion;
+using hpp::fcl::GJKConvergenceCriterionType;
+using hpp::fcl::GJKSolver;
+using hpp::fcl::ShapeBase;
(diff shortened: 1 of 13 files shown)


test/gjk_convergence_criterion.cpp
```diff
@@ -0,0 +1,169 @@
+/*
+ *  Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2022, CNRS-LAAS INRIA
```
jcarpent, line 4: You can remove LAAS-CNRS from here.

jcarpent (approved):
Looks good to me. Very nice work @lmontaut!

PR #304, opened 2022-06-05
Fix assert precision in GJK

jcarpent:
Some rounding may affect the precision.


jcarpent:
This PR handles cases where the results lead to numerical rounding issues of the form: `3e-18 <= 0` is evaluated to false, while the numerical precision of double is around `2e-16`.

PR #311, opened 2022-06-23
Fix incoherent overlap

jcarpent:
According to the definition of overlap, R0,T0 expressed the placement of the BV b1 and b2 is at the identity.
Consequently, we need either to transfert b1 to the origin or to express b2 in the frame of b1, defined by (R0,T0).


jmirabel:
It seems you adjusted the code to the documentation, and not the opposite. I am fine with this change but I am surprised you didn't do the opposite.
Is there a reason for this ?

jcarpent:
I've just aligned wrt the doc and the AABB case.

PR #314, opened 2022-06-28
Expose shared_ptr<OcTree>

Jiayuan-Gu:
Add `shared_ptr<OcTree>`, otherwise there will be `No to_python (by-value) converter found for C++ type` error for python binding `makeOcTree`

### DIFF
--- a/python/octree.cc
+++ b/python/octree.cc
@@ -13,7 +13,7 @@ void exposeOctree() {
   namespace bp = boost::python;
   namespace dv = doxygen::visitor;
 
-  bp::class_<OcTree, bp::bases<CollisionGeometry> >(
+  bp::class_<OcTree, bp::bases<CollisionGeometry>, shared_ptr<OcTree> >(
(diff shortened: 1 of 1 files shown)


nim65s:
Thanks !

Could you add a regression test, reproducing this `No to_python (by-value) converter found for C++ type` error without your fix ?

Jiayuan-Gu:
Em, I guess you need an example to reproduce. Here is a simple script to reproduce.

Install `hpp-fcl=2.0.1`
```bash
conda create -n test-hpp-fcl python=3.8
conda install hpp-fcl=2.0.1 -c conda-forge
```

And run
```python
import hppfcl as fcl
import numpy as np
fcl.makeOctree(np.random.rand(5, 3), 0.001)
```

jcarpent (approved):

jcarpent:
This PR looks good to me, as all other FCL primitives already use shared_ptr as Python container. Thanks @Jiayuan-Gu, your help is really appreciated.

PR #315, opened 2022-06-30
test/gjk_convergence_criterion: Add check on GJK::Status

wxmerkt:
We are setting but never verifying the status in the test


jcarpent (approved):
LGTM. Thanks a lot @wxmerkt for your contribution.

PR #319, opened 2022-07-14
cmake: relocatable package for recent CMake versions

nim65s:
Hi,

This follows https://github.com/stack-of-tasks/eigenpy/pull/291

It also require recent version of dependencies like assimp & octomap to unlock everything, but the general idea is the same.

There are some copy/paste from eigenpy on the python logic, while I think we could make eigenpy export everything on that matter, and ensure its dependencies would work without having to deal with python at all here, but this is for another PR.


src/CMakeLists.txt
```diff
@@ -173,20 +173,16 @@ add_library(${LIBRARY_NAME}
   ${${LIBRARY_NAME}_SOURCES}
   )
 
+set_target_properties(${PROJECT_NAME} PROPERTIES INSTALL_RPATH "\$ORIGIN")
```
jcarpent, line 176: Why this is needed?

nim65s: When we have an `install_prefix/lib` directory that gets moved around, chances are that we are going from a directory that was configured in `/etc/ld.so.conf` /  `$LD_LIBRARY_PATH` to a directory that is not. In that case, when we move `libhpp-fcl.so` and its dependencies, like eg. `libassimp.so`, the dynamic link from `libhpp-fcl.so` to `libassimp.so` is broken, as `libassimp.so` is no longer available in a `$LD_LIBRARY_PATH` dir, and the system doesn't know it should simply look into the same folder to find it.

This line therefore allows users to move the install folder around without breaking the good behaviour of the final application.

This is especially useful for python virtual environments, as thanks to that line, when the user activate an environment with assimp & hpp-fcl, he doesn't have to deal with anything else (like `$LD_LIBRARY_PATH`) to see everything working as expected out of the box.

jcarpent: Thanks for the clarification.

PR #320, opened 2022-07-15
Benchmark for Nesterov-accelerated GJK vs. GJK timings [WIP]

lmontaut:
Quick benchmark that allows to compare the timings of Nesterov-accelerated GJK against GJK. 
The goal is to see the results over multiple pairs and multiple random poses between each pair. It is also important to divide the results based on the distance between the shapes as Nesterov-accelerated GJK improves upon GJK when the problem is harder to solve for GJK (shapes in close-proximity/overlapping).

I created a dir `benchmark/meshes` where I put meshes from the panda robot found in [example-robot-data](https://github.com/Gepetto/example-robot-data) (both `collision` and `visual` meshes. I took the first 3 links of panda and its end-effector.
Is it ok to do so? 
If not, is it ok to add [example-robot-data](https://github.com/Gepetto/example-robot-data) as an optional dependency which would be needed for this benchmark? In any case, I think having a set of robot meshes in HPPFCL is a good idea as it can be reused for other benchmarks/tests.

This is still a work in progress, the code needs to be cleaned/improved. I still need to loop through all the shapes for example, but I want to be sure how to add the meshes to hppfcl first.


nim65s:
Meshes are pretty heavy, so I think it's better to keep them in another repo.

A dependency could work, but maybe a CMake `ExternalProject` that automatically trigger the download of example-robot-data when needed could be nice I guess.

jcarpent (commented):
It would be nice to move the benchmark to a top-level directory called benchmark as in Pinocchio.

jcarpent:
@lmontaut What is the status of this PR?

Issue #144, opened 2020-03-22
Returning several contact points for basic Shape collision


jcarpent:
In some cases like simulation, it would be essential to return all the collision points between two basic shapes. For instance, a box against an half-space in contact will return the four vertices of the box face in contact.

@jmirabel, @florent-lamiraux What do you think?

florent-lamiraux:
You need to better define what you expect to be returned. In the case you mention, the contact is a square surface, not only 4 points.

rayzchen:
If I use Python:

```
import hppfcl
import numpy as np

s1 = hppfcl.Box(2, 2, 2)
s2 = hppfcl.Box(2, 2, 2)
o1 = hppfcl.CollisionObject(s1)
o2 = hppfcl.CollisionObject(s2)
o2.setTranslation(np.array([1.9999, 0, 0]))
req = hppfcl.CollisionRequest()
req.num_max_contacts = 4
res = hppfcl.CollisionResult()
print(hppfcl.collide(o1, o2, req, res))
print(hppfcl.getContact(0).pos)
```

Why does this return only 1 contact? And why is that contact at (0.995, -1, -1) which is in the corner of the colliding surface? In my physics engine, it's causing the two boxes to rotate in an awkward way that just pushes them into each other, causing more collisions, causing a larger intersection volume in a vicious cycle. Shouldn't there be 4 contacts for each corner? Or at least 1 contact that is somewhere around (0.995, 0, 0)? Am I understanding the notion of a contact wrong?

jcarpent:
You're correct. The thing is that current GJK algorithm used for collision detection between two boxes only returns one contact point, not several.
The contact point information corresponds to the average of the two witness points resulting from the collision detection algorithm.

rayzchen:
How would I get a more accurate point of collision then?

jmirabel:
You must use the penetration information. Make distance requests and make sure the nearest point calculations are enabled.

rayzchen:
```python3
import hppfcl
import numpy as np

s1 = hppfcl.Box(2, 2, 2)
s2 = hppfcl.Box(2, 2, 2)
o1 = hppfcl.CollisionObject(s1)
o2 = hppfcl.CollisionObject(s2)
o1.setTranslation(np.array([1.9999999, 0, 0]))
req = hppfcl.DistanceRequest()
req.enable_nearest_points = True
res = hppfcl.DistanceResult()
hppfcl.distance(o1, o2, req, res)
[...]

PR #325, opened 2022-07-27
core: add EPA::FallBack condition to shapeDistance computation

lmontaut:
When there is a vertex-vertex collision, EPA goes into fallback mode because GJK gave it a simplex of rank 1. This corresponds to the origin being on a vertex of the Minkowski difference. In such a case, the distance is well defined (it is 0, where currently `shapeDistance` returns -infinity, see the example in #144). However, there is still some work to be done to properly define the normal.

### DIFF
--- a/include/hpp/fcl/narrowphase/gjk.h
+++ b/include/hpp/fcl/narrowphase/gjk.h
@@ -254,6 +254,9 @@ struct HPP_FCL_DLLAPI GJK {
   /// @brief Get GJK number of iterations.
   inline size_t getIterations() { return iterations; }
 
+  /// @brief Get GJK tolerance.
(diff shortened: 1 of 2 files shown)


include/hpp/fcl/narrowphase/gjk.h
```diff
@@ -254,6 +254,8 @@ struct HPP_FCL_DLLAPI GJK {
   /// @brief Get GJK number of iterations.
   inline size_t getIterations() { return iterations; }
 
+  FCL_REAL tolerance;
```
lmontaut, line 257: Should I create a `getTolerance()` method in GJK or simply move `tolerance` to public like done here?

jcarpent (approved):

PR #326, opened 2022-07-27
sync submodule

nim65s:


jcarpent (approved):

nim65s:
NB: we are now installing `python.cmake` in eigenpy, but the `cmake/python.cmake` in hpp-fcl takes precedence, so if it's not up-to-date we might get unexpected behaviours. I don't think we currently have an issue here, but I think it's better to synchronize now.

nim65s:
Actually… we can fix this issue at the source.

PR #327, opened 2022-07-27
 CMake: update to eigenpy 2.7.10

nim65s:
This will have to wait for eigenpy 2.7.10 to be available everywhere before the CI can succeed.

### DIFF
--- a/.github/workflows/conda/conda-env.yml
+++ b/.github/workflows/conda/conda-env.yml
@@ -8,7 +8,7 @@ dependencies:
   - assimp
   - numpy
   - boost
-  - eigenpy=2.7.6
+  - eigenpy
   - python
   - doxygen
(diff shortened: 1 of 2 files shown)


nim65s:
I thought we had to wait for eigenpy 2.7.10 being available in apt, but actually just restarting the ROS jobs did the work.

wxmerkt:
> just restarting the ROS jobs did the work.

Because in the meantime the release to the buildfarm happened and now 2.7.10 is available on staging (but not release yet until the sync - so unless users have ros-testing instead of ros-main they won't get 2.7.10 just yet)

wxmerkt:
Can we have a release tagged with this update please?

jcarpent:
I'm doing that now ;)

Issue #307, opened 2022-06-16
Discrepancy btw GJK+EPA and specialized functions regarding witness points


lmontaut:
Hi,
I have the following script which computes the distance between a sphere and a box:
```
import hppfcl
import numpy as np

r = 0.5
o1 = hppfcl.Sphere(r)
# o1 = hppfcl.Ellipsoid(r, r, r)
T1 = hppfcl.Transform3f(np.eye(3), np.array([0, 0, r]))

rbox = 2 * r
o2 = hppfcl.Box(rbox, rbox, rbox)
T2 = hppfcl.Transform3f(np.eye(3), np.zeros(3))

request = hppfcl.DistanceRequest()
result = hppfcl.DistanceResult()
dist = hppfcl.distance(o1, T1, o2, T2, request, result)
print(f"Distance = {dist}")
print(f"Normal = {result.normal}")
print(f"p1 = {result.getNearestPoint1()}")
print(f"p2 = {result.getNearestPoint2()}")
```
which outputs:
```
Distance = -0.5
Normal = [-0. -0. -1.]
p1 = [0.  0.  0.5]
p2 = [0.  0.  0.5]
```
Under the hood, `hppfcl.distance(...)` is using the **specialized function** `sphereSphereDistance(...)`.

If I now replace `o1 = hppfcl.Sphere(r)` with `o1 = hppfcl.Ellipsoid(r, r, r)`, `hppfcl.distance(...)` is now using **GJK+EPA**.
I get the following output:
```
Distance = -0.4999999999989329
Normal = [-1.16274225e-11  1.18861597e-11 -1.00000000e+00]
p1 = [ 5.43733207e-07 -4.25368667e-07  1.06714637e-12]
p2 = [ 5.43739021e-07 -4.25374610e-07  5.00000000e-01]
```
I essentially get the same result *except* for the position of the witness points `p1` and `p2`. This example illustrates the fact that there seems to be a discrepancy between how the specialized collision/distance functions and GJK+EPA view the nearest points when there is a collision. It seems that most specialized functions (located in `narrowphase/detail.h`) see the nearest points as sharing the same position, whereas GJK+EPA returns the witness points which correspond to the separation vector.

Is this intentional? In my opinion, the output of GJK+EPA is the correct one as the points it outputs are where contact forces would be applied on each shape.

jmirabel:
Originally, the notion of contact point was very fuzzy in FCL. I think it is any point in the intersection.

[...]

PR #329, opened 2022-07-29
Normal and nearest points uniformization

lmontaut:
This PR addresses #307.

In short:
The library is made uniform w.r.t the definition of the witness points (in hppfcl they are called `nearest_points`, we might want to change that to avoid confusion) and the normal. 

In details:
Depending on which collision pair between two objects `o1` and `o2`, there were concurrent definitions of their normal and their witness points. The reason is that GJK and EPA were rewritten in hppfcl but followed a different convention compared to the original fcl library when it comes to the definition of the normal. 
Thus, collision pairs involving specialized functions (not using GJK+EPA) were behaving differently than GJK+EPA. 

In addition, the specialized functions returned collocated witness points whereas GJK+EPA returned the true witness points.

To address this, this PR does the following:
- adds documentation to define the normal and the nearest points. In short, it simply expands on the documentation already present in hppfcl: the normal always points from `o1` to `o2`. It follows that the distance `d(o1, o2)` between the shapes is always a **signed** distance (even when using `collide`, contrary to current behavior): it is negative when collision, positive otherwise. The normal `n` and the witness points `p1` and `p2` are related in that the normal always points in the direction of the separation vector: `separation_vector = sign(d(o1, o2)) * (p2 - p1) and `n = separation_vector.normalized()`. The separation vector can be used to bring the shapes into touching contact (share at least a contact point bu zero intersection volume). Indeed, if we translate `o1` by `separation_vector`, regardless of collision or not, we bring the shapes into touching contact.
- adds a test to make sure all collision pairs' normal and nearest points verify the definition stated above. This checks tests the relationship between the normal, the separation vector, the witness points, the signed distance etc. Once this is passed, it translates `o1` according to the separation vector and makes sure that the shapes are in collision if they were not or not in collision if they were.
- Updates the behavior of GJK+EPA: now the normal always points from o1 to o2. Previously GJK+EPA's normal was pointing from o2 to o1 when the shapes were colliding.
- Updates the behavior of all the specialized functions accordingly.

This PR is long, apologies for that. I think that the most important is to understand and agree on the definition of the normal given above. I sticked to the existing documentation, i'm happy to update it if everyone agrees on a different definition of the normal. The other important thing is to understand the test I added, `test/normal_and_nearest_points.cpp`, which guarantees all collision pairs follow the same definition. 

All of this corresponds to the first 9 commits of the PR, the rest is simply updating each specialized function.


include/hpp/fcl/internal/shape_shape_func.h
```diff
@@ -136,6 +135,11 @@ SHAPE_SHAPE_DISTANCE_SPECIALIZATION(Sphere, Halfspace);
 SHAPE_SHAPE_DISTANCE_SPECIALIZATION(Sphere, Plane);
 SHAPE_SHAPE_DISTANCE_SPECIALIZATION(Sphere, Sphere);
 SHAPE_SHAPE_DISTANCE_SPECIALIZATION(Sphere, Cylinder);
+// TODO
+// SHAPE_SHAPE_DISTANCE_SPECIALIZATION(Sphere, Capsule);
+// SHAPE_SHAPE_DISTANCE_SPECIALIZATION(TriangleP, TriangleP);
+// SHAPE_SHAPE_DISTANCE_SPECIALIZATION(Ellispoids, Plane);
+// SHAPE_SHAPE_DISTANCE_SPECIALIZATION(Ellispoids, Halfspace);
```
jcarpent, line 142: @lmontaut You still have some todos. Is it on purpose?

python/collision.cc
```diff
@@ -59,6 +59,15 @@ const CollisionGeometry* geto(const Contact& c) {
...
 
+struct ContactWrapper {
+  static Vec3f getNearestPoint1(const Contact& contact) {
+    return contact.nearest_points[0];
+  }
+  static Vec3f getNearestPoint2(const Contact& contact) {
+    return contact.nearest_points[1];
+  }
+};
+
```
jcarpent, line 70: I guess you add some difficulties to exposing directly `nearest_points`?

src/narrowphase/details.h
```diff
@@ -1429,37 +1424,25 @@ inline bool boxHalfspaceIntersect(const Box& s1, const Transform3f& tf1,
...
+  for (Vec3f::Index i = 0; i < 3; ++i) {
+    // scalar product between box axis and hyperplane normal
+    FCL_REAL alpha(R.col(i).dot(new_s2.n));
+    if (alpha > eps) {
+      p1 -= R.col(i) * s1.halfSide[i];
+    } else if (alpha < -eps) {
+      p1 += R.col(i) * s1.halfSide[i];
+    }
+  }
+  p2.noalias() = p1 - distance * new_s2.n;
```
jcarpent, line 1441:
```suggestion
  p2 = p1 - distance * new_s2.n;
```

jcarpent (commented):
The PR seems good to me. 
I suggest creating a new branch called hppfcl3x, as the modifications imply a break in the API.

include/hpp/fcl/internal/shape_shape_func.h
lmontaut: Yes. I will do these in the next PR because it will add a lot of lines to the current one, so I want to avoid confusions.

python/collision.cc
lmontaut: I followed what was proposed in order to expose `nearest_points` in `DistanceResult`. If you have a better idea, I'll take it :)

jcarpent: We should also expose the `nearest_points` quantity to be more complete.

lmontaut: I agree but when I try to expose `nearest_points` in `Contact` or `DistanceResult`, via `.DEF_RW_CLASS_ATTRIB(Contact, b1)`, eigen is not happy:
```
In file included from /Users/louis/code/control/hpp-fcl.git/normal_and_nearest_points/python/collision.cc:35:
In file included from /Users/louis/anaconda3/envs/hppfcl/include/eigenpy/eigenpy.hpp:10:
In file included from /Users/louis/anaconda3/envs/hppfcl/include/eigenpy/eigen-typedef.hpp:8:
In file included from /Users/louis/anaconda3/envs/hppfcl/include/eigenpy/fwd.hpp:14:
In file included from /Users/louis/anaconda3/envs/hppfcl/include/boost/python.hpp:18:
In file included from /Users/louis/anaconda3/envs/hppfcl/include/boost/python/class.hpp:17:
/Users/louis/anaconda3/envs/hppfcl/include/boost/python/data_members.hpp:57:22: error: array type 'Eigen::Matrix<double, 3, 1, 0> [2]' is not assignable
          c.*m_which = d;
```

jcarpent: In fact, it is Boost.Python which is not happy. It is not able to expose C array.
The solution could consist in using std::array instead of simple C array.

lmontaut: Done.

lmontaut:
The CI is failing because I didn't serialize the `std::array` properly.

nim65s:
The CI is also failing because #327. Can you rebase this on devel ?

lmontaut:
> The CI is also failing because #327. Can you rebase this on devel ?

Done. Thanks @nim65s ! Is it also possible to rebase `hppfcl3x` on devel? I am not sure I have the rights to do that.

nim65s:
hppfcl3x is behind devel, so you can fast-forward by just pushing there. I think you should have the right for this. Can you check this ?

lmontaut:
Yes, I do have the right, thanks!

florent-lamiraux:
@lmontaut : Thank you for this valuable work. I did not look into all the modifications. Can you confirm that after calling `fcl::ComputeCollision::operator()`, if there is a collision, the method `CollisionResult::getContact` still returns a point that is in the intersection between the two objects ?

lmontaut:
Hi @florent-lamiraux, yes `CollisionResult::getContact` still returns a point in the intersection, no matter if `collide` or `ComputeCollision` is used. In fact, for most pairs, `getContact` returns the same point as before. The changes I made only change the contact point's position for `halfspace-...` and `plane-...`. 

A bit more details: the changes I made is to add the `nearest_points` attribute in the `Contact`struct  (used inside `CollisionResult`) such that the contact point (`CollisionResult::getContact(i).pos`) is `0.5*(nearest_points[0] + nearest_points[1]`. The `nearest_points` are defined such that their difference gives the separation vector (vector of smallest norm such that o1 translated by that vector bring o1 and o2 to touching contact).

florent-lamiraux:
Great. Thank you.

PR #331, opened 2022-08-10
Minor boost cleanup

pantor:
Two minor changes:
 - Replacing `boost::math::isnan` with `std::isnan` (which is already used by the project).
 - Removed two unused boost includes.


jcarpent (approved):
Thanks @pantor for this contribution.

nim65s (approved):
Thanks !

PR #332, opened 2022-08-17
[CI] Activate ROS2 configurations

wxmerkt:


jcarpent:
@wxmerkt Do you know what happens?

wxmerkt:
Not yet - we first need to fix EigenPy on 18.04 (the generated config since 2.7.11 no longer links against the correct Boost Python - it was fine in 2.7.10)

jcarpent (approved):

Issue #333, opened 2022-08-18
Qhull update submodule?


gianluigigrandesso:
The Qhull commit is too old (2016) and it doesn't generate the cmake config files. @ManifoldFR

ManifoldFR:
We noticed this while installing on a Ubuntu 18.04 system. Right now the CMake config for HPP expects a CMake config file for Qhull to be somewhere on the path. Qhull only generates such a file since [this commit](https://github.com/qhull/qhull/commit/925136ab82bce7fb45bf480ebc98cc68d3b88bfd)
This doesn't happen on systems where the apt package for Qhull installs the config file.

jcarpent:
You should rather install qhull via apt or any package manager.
@jmirabel Could we remove the qhull version coming as a third party?

jmirabel:
The original reason for having qhull as a submodule is that the version on apt official repo did not contain one of the required library to link against (I don't remember which one). I think the problem was reported somewhere and a fix was proposed. This was long ago (probably more than 5 years ago) so it may not be an issue anymore. If it is the case, then removing the submodule is the way to go.

nim65s:
> Right now the CMake config for HPP expects a CMake config file for Qhull to be somewhere on the path.

I think if this CMake config file isn't found, we fallback to another strategy. Could you provide the exact error you get ?

nim65s:
The submodule is here for people with a Qhull v7 from ubuntu 18.04 and 20.04 (to workaround https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=925540) so I think it has to be on the same version. 

People on ubuntu 22.04 can have a Qhull v8 from their system, and in that case CMake will find the config file, and avoid building stuff from the submodule.

jcarpent:
Fine with me. We can close this issue then.

PR #335, opened 2022-08-24
Remove QHULL from source installation

jcarpent:
Related to #333

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
 [submodule "cmake"]
 	path = cmake
 	url = https://github.com/jrl-umi3218/jrl-cmakemodules.git
-[submodule "third-parties/qhull"]
-	path = third-parties/qhull
-	url = https://github.com/qhull/qhull.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,7 @@ endif()
 
 option(HPP_FCL_HAS_QHULL "use qhull library to compute convex hulls." FALSE)
 if(HPP_FCL_HAS_QHULL)
-  ADD_PROJECT_DEPENDENCY(Qhull REQUIRED COMPONENTS qhull_r qhullcpp)
+  ADD_PROJECT_DEPENDENCY(Qhull REQUIRED COMPONENTS qhull_r qhullcpp FIND_EXTERNAL Qhull)
 endif()
 
 ADD_PROJECT_DEPENDENCY(assimp REQUIRED FIND_EXTERNAL "assimp")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 #
 # Software License Agreement (BSD License)
 #
-#  Copyright (c) 2014-2020 CNRS-LAAS, INRIA
+#  Copyright (c) 2014-2022 CNRS-LAAS, INRIA
 #  Author: Florent Lamiraux, Joseph Mirabel
 #  All rights reserved.
 #
@@ -139,7 +139,7 @@ if(HPP_FCL_HAS_QHULL)
   endif()
 endif()
 
-FIND_PACKAGE(assimp REQUIRED)
+ADD_PROJECT_DEPENDENCY(assimp REQUIRED FIND_EXTERNAL "assimp")
 
 SET(${PROJECT_NAME}_HEADERS
   include/hpp/fcl/collision_data.h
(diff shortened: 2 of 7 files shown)


nim65s (approved):

jcarpent:
@nim65s Could you check the issue on the gitlab CI?

nim65s:
We'll probably need a FindQhull.cmake

nim65s:
Well, no, this won't provide Qhull:: targets. I'll be lazy and package a more recent version of qhull on robotpkg. Links are private anyway.

jcarpent:
Seems reasonable indeed.

wxmerkt:
>  I'll be lazy and package a more recent version of qhull on robotpkg

Are there issues with the system provided QHull on 20.04 and what happens if that custom qhull from robotpkg is not available? (E.g. the ROS buildfarm so far had no issues in building HPP-FCL - would there be issues now?)

jcarpent:
We should simply provide a correct FindQhull.cmake to properly package the project. The reason is that this is not a good idea to internally package another largely used lib like qhull in hpp-fcl.

@nim65s Could we use something like: https://github.com/PointCloudLibrary/pcl/blob/master/cmake/Modules/FindQhull.cmake?

nim65s:
Qhull has cmake exports since v8, and v8 is available from Debian 11 / Ubuntu 22.04.

You're right, packaging it on robotpkg wouldn't fix ROS use cases… But why did the CI succeed then ? Maybe it simply was never enabled ?

The FindQhull.cmake from pcl could be a nice solution indeed, until  we have qhull official cmake exports available everywhere.

jcarpent:
@nim65s What is the status of the find helpers?

wxmerkt:
We can test ROS-CI now so this PR is unblocked. Once it's merged, can we tag a new release as well please? Thank you :-)

nim65s:
My latests commits are broken, I'll need to write another FindQhull.cmake first

nim65s:
Actually, it looks like we need qhullcpp, which is available in QHull 8 (2020) on ubuntu 22.04, but not in QHull 7 (2015) on ubuntu 20.04. 

In other words, I don't think we can Remove QHULL from source installation unless we abandon Ubuntu 20.04 (and our Talos is still on Ubuntu 18.04…)

If nobody has a better option, I'll update the submodule and its use to QHull 8 and; and allow the use of a system qhull only if >= 8 is detected.

jcarpent:
Sounds good to me. Thanks @nim65s for checking it further.

nim65s:
1. Providing a recent Qhull submodule with the current strategy doesn't work, because we find ubuntu's `libqhull_r.so` v7, so we can't just build `libqhullcpp.a` v8.

2. Providing a recent Qhull submodule with `add_subdirectory()`doesn't work, because Qhull provides a `uninstall` CMake global targets which collides with ours from jrl-cmakemodule

3. Providing a recent Qhull via `FetchContent_Declare()` won't work on 18.04, because its default CMake is 3.10, and FetchContent is available since 3.11.

4. Providing a recent Qhull via `ExternalProject_add()` doesn't allow us to get the include directories at config time, so we can't specify those.

5. Providing a recent Qhull submodule to get include directories in the source tree + `ExternalProject_add()` with a local path as `GIT_REPOSITORY` provides everything we need, except `libqhull_r.so` is shared, so we would have to ship it with hpp-fcl binaries, which wouldn't be an excellent idea. 

6. I didn't manage to correctly link `libhpp-fcl.so` to the static version of that lib, `libqhullstatic_r.a`, because this would require everything linking to `libhpp-fcl.so` to be compiled with `-fPIC`, and that's not the case for eg. ubuntu's `libassimp.so`.

So I think we shouldn't do anything here. I also think that #333 is misleading, and that we can fix it without changing anything in this repo. If we really want to do something, 5 seems the only plausible solution, but we'll have to ensure that the `libqhull_r.so` we'll ship in our binaries won't interfere with anything. In that case would suggest to rename it to `libhpp-fcl-qhull-r.so` or something.

jcarpent:
> If we really want to do something, 5 seems the only plausible solution, but we'll have to ensure that the `libqhull_r.so` we'll ship in our binaries won't interfere with anything. In that case would suggest to rename it to `libhpp-fcl-qhull-r.so` or something.

I think this is the best solution I would say. But requires some work.
What would be the work to be compliant with QHull 7 @jmirabel? Do you know which features are not present?

jmirabel:
The issues @nim65s mention look very similar to the issues that lead me to include QHull into this repo.

jmirabel:
For reference, QHull was integrated in https://github.com/humanoid-path-planner/hpp-fcl/pull/167 and the rationale of what I did is roughly explained in the PR.

I think we should:
- 18.04: not change the current behavior
- 20.04: I'm not sure. I haven't checked what the ubuntu repo provides
- 22.04: rely only on qhull from the official repo.

Is this feasible ?

nim65s:
I agree. And for this, I think all we have to do is keep the repo in its current state.

jcarpent:
I do agree with your conclusions. My only worry is that we have to cheap QHULL with hpp-fcl for certain OS, which is a bit weird IMHO.

Issue #338, opened 2022-09-07
More flexible CMake configuration


JeffHough:
I had some trouble configuring the cmake for this project. Specifically, I'll list out a few problems I ran into that seem like they are potentially caused by how the `CMakeLists.txt` is written, but I am not 100% sure.

## Is it possible to _not_ build with octomap?

As seen in the cmake:

```
# Optional dependencies
ADD_PROJECT_DEPENDENCY(octomap PKG_CONFIG_REQUIRES "octomap >= 1.6")
if(octomap_FOUND)
  SET(HPP_FCL_HAS_OCTOMAP TRUE)
  string(REPLACE "." ";" VERSION_LIST ${octomap_VERSION})
  list(GET VERSION_LIST 0 OCTOMAP_MAJOR_VERSION)
  list(GET VERSION_LIST 1 OCTOMAP_MINOR_VERSION)
  list(GET VERSION_LIST 2 OCTOMAP_PATCH_VERSION)
  message(STATUS "FCL uses Octomap")
else()
  SET(HPP_FCL_HAS_OCTOMAP FALSE)
  message(STATUS "FCL does not use Octomap")
endif()
```

Which, it would appear, forces octomap to be included if it can be found. If octomap is an optional dependency, should it not be an `option`? Something like: 

```
option(HPP_FCL_USE_OCTOMAP "whether to use octomap" OFF)
# Optional dependencies
if (HPP_FCL_USE_OCTOMAP ON)
  ADD_PROJECT_DEPENDENCY(octomap PKG_CONFIG_REQUIRES "octomap >= 1.6")
  if(octomap_FOUND)
    SET(HPP_FCL_HAS_OCTOMAP TRUE)
    string(REPLACE "." ";" VERSION_LIST ${octomap_VERSION})
    list(GET VERSION_LIST 0 OCTOMAP_MAJOR_VERSION)
    list(GET VERSION_LIST 1 OCTOMAP_MINOR_VERSION)
    list(GET VERSION_LIST 2 OCTOMAP_PATCH_VERSION)
    message(STATUS "FCL uses Octomap")
  else()
    SET(HPP_FCL_HAS_OCTOMAP FALSE)
    message(STATUS "FCL does not use Octomap")
  endif()
endif()
```
That way, I could simultaneously have a system install of `octomap` and still choose not to use it?

[...]

PR #350, opened 2022-09-15
Updated the cmake to be more configurable [WIP]

JeffHough:
This PR is in response to issue #338.

I've added some options in the root `CMakeLists.txt` file, and set the default options so that things should be the same as they are now if no flags are set.

These are the new options:
```
OPTION(HPP_FCL_USE_OCTOMAP "Build hpp-fcl with octomap" ON)
OPTION(HPP_FCL_BUILD_TESTS "Build the tests or not" ON)
OPTION(HPP_FCL_BUILD_CORE_ONLY "Only build the core functionality of hpp-fcl (no mesh loaders, no ROS, no tests)" OFF)
OPTION(HPP_FCL_BUILD_SHARED_LIBRARY "Build hpp-fcl as a shared library. Otherwise static." ON)
```

All of the flags are pretty much what they sound like. 

For the `HPP_FCL_BUILD_CORE_ONLY`, this will not build anything related to catkin or colcon, and will also not include any mesh loaders (i.e. no longer requires `assimp`) and forces the tests to be off (the tests fail to link without assimp so cannot build). I added this flag for my own particular use case where I am embedding `hpp-fcl` in another application for collision-detection, and my mesh files are already loaded. That is, I do not need a mesh loader, and would prefer to not have the additional dependency.


CMakeLists.txt
```diff
@@ -52,6 +52,24 @@ SET(DOXYGEN_USE_TEMPLATE_CSS TRUE)
 # Need to be set before including base.cmake
 # ----------------------------------------------------
 OPTION(INSTALL_DOCUMENTATION "Generate and install the documentation" OFF)
+OPTION(HPP_FCL_USE_OCTOMAP "Build hpp-fcl with octomap" ON)
+OPTION(HPP_FCL_BUILD_TESTS "Build the tests or not" ON)
```
jcarpent, line 56: This is redundant with `BUILD_TESTING`. Could you remove it?

CMakeLists.txt
```diff
@@ -52,6 +52,24 @@ SET(DOXYGEN_USE_TEMPLATE_CSS TRUE)
...
+IF (HPP_FCL_BUILD_CORE_ONLY)
+  SET(INSTALL_DOCUMENTATION OFF)
+
+  # tests fail to link without assimp.
+  SET(HPP_FCL_BUILD_TESTS OFF)
+  MESSAGE(WARNING "BUILDING CORE ONLY - TESTS ARE OFF.")
+ENDIF()
+
+IF (HPP_FCL_BUILD_SHARED_LIBRARY)
+  SET(HPP_FCL_BUILD_TYPE SHARED)
```
jcarpent, line 69: This should be rather named `HPP_FCL_LIBRARY_TYPE`

CMakeLists.txt
```diff
@@ -152,8 +175,6 @@ if(HPP_FCL_HAS_QHULL)
   endif()
 endif()
 
-FIND_PACKAGE(assimp REQUIRED)
```
jcarpent, line 155: We can make assimp as an optional dependency. Could you add it rather than using `HPP_FCL_BUILD_CORE_ONLY` which is a not a good practice.

CMakeLists.txt
```diff
@@ -259,12 +278,24 @@ SET(${PROJECT_NAME}_HEADERS
...
+ENDIF()
+
 add_subdirectory(src)
+
 if (BUILD_PYTHON_INTERFACE)
   add_subdirectory(python)
 endif ()
-add_subdirectory(test)
+
+IF (HPP_FCL_BUILD_TESTS)
```
jcarpent, line 296: We should use BUILD_TESTING as default. Could you remove it?

CMakeLists.txt
```diff
@@ -273,12 +304,14 @@ IF(HPP_FCL_HAS_OCTOMAP)
...
+  # Install catkin package.xml
+  INSTALL(FILES package.xml DESTINATION share/${PROJECT_NAME})
+  # Allows Colcon to find non-Ament packages when using workspace underlays
+  file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/share/ament_index/resource_index/packages/${PROJECT_NAME} "")
+  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/share/ament_index/resource_index/packages/${PROJECT_NAME} DESTINATION share/ament_index/resource_index/packages)
+  file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/share/${PROJECT_NAME}/hook/ament_prefix_path.dsv "prepend-non-duplicate;AMENT_PREFIX_PATH;")
+  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/share/${PROJECT_NAME}/hook/ament_prefix_path.dsv DESTINATION share/${PROJECT_NAME}/hook)
+  file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/share/${PROJECT_NAME}/hook/python_path.dsv "prepend-non-duplicate;PYTHONPATH;${PYTHON_SITELIB}")
+  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/share/${PROJECT_NAME}/hook/python_path.dsv DESTINATION share/${PROJECT_NAME}/hook)
+ENDIF()
```
jcarpent, line 317: You should not change that. This installation.

CMakeLists.txt
```diff
@@ -52,6 +52,24 @@ SET(DOXYGEN_USE_TEMPLATE_CSS TRUE)
...
+OPTION(HPP_FCL_BUILD_CORE_ONLY "Only build the core functionality of hpp-fcl (no mesh loaders, no ROS, no tests)" OFF)
+OPTION(HPP_FCL_BUILD_SHARED_LIBRARY "Build hpp-fcl as a shared library. Otherwise static." ON)
+
+IF (HPP_FCL_BUILD_CORE_ONLY)
+  SET(INSTALL_DOCUMENTATION OFF)
+
+  # tests fail to link without assimp.
+  SET(HPP_FCL_BUILD_TESTS OFF)
+  MESSAGE(WARNING "BUILDING CORE ONLY - TESTS ARE OFF.")
+ENDIF()
```
jcarpent, line 66: This setting is very specific to your own setting.
I prefer to make assimp optional, and to properly do it rather than trying to add specific build core which hard to maintain in the end

jcarpent (changes requested):
Thanks for this contribution. I made major comments which should be addressed.
In general, I think the way you implement your solution is very specific.

I don't think the introduction of `HPP_FCL_BUILD_CORE_ONLY` concept is really useful. This is not a clean solution. I made some suggestions to converge towards a better setup.
@nim65s Could you provide more feedback?

JeffHough:
@jcarpent thanks for all the feedback! 

As you're saying, this is basically a configuration that I found to work well for me, but I agree, the optional dependency of `assimp`, if wanted to be optional at all, could probably be done in a more general way.

When I have a bit of time I'll make the edits you recommend!

CMakeLists.txt
```diff
@@ -52,6 +52,24 @@ SET(DOXYGEN_USE_TEMPLATE_CSS TRUE)
 # Need to be set before including base.cmake
 # ----------------------------------------------------
 OPTION(INSTALL_DOCUMENTATION "Generate and install the documentation" OFF)
+OPTION(HPP_FCL_USE_OCTOMAP "Build hpp-fcl with octomap" ON)
+OPTION(HPP_FCL_BUILD_TESTS "Build the tests or not" ON)
+OPTION(HPP_FCL_BUILD_CORE_ONLY "Only build the core functionality of hpp-fcl (no mesh loaders, no ROS, no tests)" OFF)
+OPTION(HPP_FCL_BUILD_SHARED_LIBRARY "Build hpp-fcl as a shared library. Otherwise static." ON)
```
nim65s, line 58: I would be in favor of putting this option in the jrl-cmakemodules, to allow other projects to benefit from this.
But we can go ahead with this PR and merge it like this for now, if you are OK to rename this in a more general way later.

nim65s (commented):
I agree with @jcarpent, and added a minor remark

PR #360, opened 2022-11-25
Extract checks if AABB overlap

jmirabel:


src/collision_utility.cpp
```diff
@@ -21,13 +21,25 @@
...
 namespace details {
+
 template <typename NT>
 inline CollisionGeometry* extractBVHtpl(const CollisionGeometry* model,
                                         const Transform3f& pose,
                                         const AABB& aabb) {
+  // Ensure AABB is already computed
+  if (model->aabb_radius < 0)
+    throw std::invalid_argument(
+        "Collision geometry AABB should be computed first.");
```
jcarpent, line 32:
```suggestion
  if (model->aabb_radius < 0)
    HPP_FCL_THROW_PRETTY("Collision geometry AABB should be computed first.",std::invalid_argument());
```

jcarpent (changes requested):
Thanks @jmirabel for the PR.
I propose using `HPP_FCL_THROW_PRETTY` to uniformize the throw within hpp-fcl.

jcarpent (approved):

PR #371, opened 2023-01-11
Adding EarlyStopped flag in GJK

lmontaut:
When GJK early stops, the closest points `p1` and `p2` make no sense, so we can simply set them to `nan`.

Q: What should we do about `distance`? For now it is set to the distance between the origin and the separating hyperplane, but we could set it to `nan` like the closest points.


jcarpent:
Here the distance has a meaning as a lower distance bound.

jcarpent (approved):
Small comment: the idea of using NaN values is for propagation in the computational graph. For the end-user, no contact should be added in case a separating hyperplane is found.

jcarpent (approved):

jcarpent:
Importantly, this PR solves a bug occurring in Pinocchio in debug mode.

PR #378, opened 2023-02-16
Skip test if BUILD_TESTING is OFF

jcarpent:


nim65s:
If we add this check here, we should remove it from the deeper level, eg: 
https://github.com/humanoid-path-planner/hpp-fcl/blob/00982e74f034bbb489e0cef2142bc4b94ab331bd/test/CMakeLists.txt#L26

jcarpent:
Thanks @nim65s for the quick feedback. I've handled the remark.

PR #381, opened 2023-03-13
Fix issues on normal directions for hfields

jcarpent:


include/hpp/fcl/collision_data.h
```diff
@@ -134,6 +134,8 @@ struct HPP_FCL_DLLAPI Contact {
   }
 
   bool operator!=(const Contact& other) const { return !(*this == other); }
+
+  inline FCL_REAL getDistanceToCollision(const CollisionRequest& request);
```
jcarpent, line 138:
```suggestion
  FCL_REAL getDistanceToCollision(const CollisionRequest& request) const;
```

PR #385, opened 2023-03-19
Remove useless call to /proc/cpuinfo

### LINKED ISSUE #384: Counting processors
Hi,

The release v2.3.0 on cmeel is currently failing on MacOS Intel at:
```
        15/33 Test #15: obb ..............................***Exception: SegFault  0.11 sec
        failed to open /proc/cpuinfo
```

I don't see why this is hitting now and did not occur before, but anyways: is reading `/proc/cpuinfo` really the right way on Mac OS and Windows ?

Otherwise, I guess we could rely on [CMake `ProcessorCount`](https://cmake.org/cmake/help/latest/module/ProcessorCount.html) to achieve the same goal.

jcarpent:
Resolves #384

### DIFF
--- a/test/obb.cpp
+++ b/test/obb.cpp
@@ -2,6 +2,7 @@
  * Software License Agreement (BSD License)
  *
  *  Copyright (c) 2014-2016, CNRS-LAAS
+ *  Copyright (c) 2023, Inria
  *  Author: Florent Lamiraux
  *  All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


test/obb.cpp
```diff
@@ -1378,16 +1316,6 @@ int main(int argc, char** argv) {
...
 
-  bool cpuScalingEnabled = checkCpuScalingEnabled();
-  if (cpuScalingEnabled)
-    std::cerr << "CPU scaling is enabled."
-                 "\n\tThe benchmark real time measurements may be noisy and "
-                 "will incur extra overhead."
-                 "\n\tUse the following commands to turn on and off."
-                 "\n\t\tsudo cpufreq-set --governor performance"
-                 "\n\t\tsudo cpufreq-set --governor powersave"
-                 "\n";
```
jmirabel, line 1389: I'm fine removing the check whether CPU scaling is enabled. However, I would print a message to `std:::cout` similar to the one above. It's very easy not to pay attention to this.

jcarpent: I've restored it. Thanks @jmirabel for the feedback.

jmirabel (approved):

PR #387, opened 2023-03-24
[GJKSolver] Fix missing switch case in result status of GJK

lmontaut:


jcarpent (approved):

include/hpp/fcl/narrowphase/narrowphase.h
```diff
@@ -250,7 +251,7 @@ struct HPP_FCL_DLLAPI GJKSolver {
         break;
       default:
         assert(false && "should not reach type part.");
-        return true;
+        throw std::logic_error("GJKSolver: should not reach this part.");
```
jmirabel, line 254: The assert just before should be removed I think.

PR #393, opened 2023-03-31
Sync submodule cmake

jcarpent:


nim65s (approved):

nim65s:
I wanted to 2 things at the same time, but the second is broken, so let's keep to the first and move the second to #394

nim65s (approved):

PR #410, opened 2023-04-28
update default C++ to 14

nim65s:
to avoid hundreds of boost 1.81 warnings:
```
#warning "The minimum language standard to use Boost.Math will be C++14 starting in July 2023 (Boost 1.82 release)"
```

Also, there shouldn't be much use cases where C++11 is available and C++14 is not.


jcarpent (commented):
It is a big jump

jcarpent (changes requested):
Could we make it dependent on the Boost version to not force to much if it is not necessary?

jcarpent:
@wxmerkt It seems the ROS CI is failing (similarly to Pinocchio). Could you help us on it?

CMakeLists.txt
```diff
@@ -111,6 +108,14 @@ if(BUILD_PYTHON_INTERFACE)
   find_package(Boost REQUIRED COMPONENTS system)
 endif(BUILD_PYTHON_INTERFACE)
 
+if(Boost_VERSION_STRING VERSION_LESS 1.82)
```
nim65s, line 111:
```suggestion
if(Boost_VERSION_STRING VERSION_LESS 1.81)
```

This would allow us to test it with Boost 1.81 and get rid of the warnings

nim65s:
I think we could merge this, and fix ROS-CI in another PR, especially because this other issues hits other repositories

wxmerkt:
The problems with ROS-CI were due to unavailability of the EigenPy binary on Rolling - this issue was temporary (RAM exhaustion) and has been resolved. I restarted the jobs so hopefully they should come back green now

jcarpent (approved):

PR #420, opened 2023-05-11
Modernize Cmake use

nim65s:


jcarpent (changes requested):
Could you please split the PR into two parts? I’m not able to review between the precommit and the core change

jcarpent (approved):

PR #432, opened 2023-06-01
octree.cpp: add missing include

ManifoldFR:


nim65s (approved):
You're aware this is also adding `test/nesterov_gjk.cpp`, right ? :)

ManifoldFR:
> You're aware this is also adding `test/nesterov_gjk.cpp`, right ? :)

Oops, that's what I get for doing a `git add .` without checking...

jcarpent (changes requested):
@ManifoldFR I will make a new release of hppfcl 2x and I will update afterward hppfcl 3x. It will simplify the sync between the two branches. OK?

jcarpent (approved):

Issue #446, opened 2023-07-20
How should I cite this repo?


BolunDai0216:
Is it possible to provide the BibTeX for citing this repo? 

It would be easier for others if you can include a CITATION.cff file in this repo.

jcarpent:
Thanks for citing this repository.
You can use this citation entry for citing the code:
```bib
@misc{hppfclweb,
   author = {Jia Pan and Sachin Chitta and Dinesh Manocha and Florent Lamiraux and Joseph Mirabel and Justin Carpentier and Louis Montaut and others},
   title = {HPP-FCL: an extension of the Flexible Collision Library},
   howpublished = {https://github.com/humanoid-path-planner/hpp-fcl},
   year = {2015--2023}
}
```
and this one for the acceleration of collision detection:
```bib
@INPROCEEDINGS{Montaut-RSS-22, 
    AUTHOR    = {Louis Montaut AND {Quentin Le} Lidec AND Vladimír Petrík AND Josef Sivic AND Justin Carpentier}, 
    TITLE     = {{Collision Detection Accelerated: An Optimization Perspective}}, 
    BOOKTITLE = {Proceedings of Robotics: Science and Systems}, 
    YEAR      = {2022}, 
    ADDRESS   = {New York City, NY, USA}, 
    MONTH     = {June}, 
    DOI       = {10.15607/RSS.2022.XVIII.039} 
} 
```

I've just opened a PR (#449) on this topic.

PR #449, opened 2023-07-24
Add citations

jcarpent:
Related to #446

### DIFF
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -1,0 +1,6 @@
+@misc{hppfclweb,
+   author = {Jia Pan and Sachin Chitta and Dinesh Manocha and Florent Lamiraux and Joseph Mirabel and Justin Carpentier and Louis Montaut and others},
+   title = {HPP-FCL: an extension of the Flexible Collision Library},
+   howpublished = {https://github.com/humanoid-path-planner/hpp-fcl},
+   year = {2015--2023}
+}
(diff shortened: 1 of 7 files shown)


nim65s (approved):

jcarpent:
Currently, we have some issues with Eigen 3.4.1 as there is compilation bugs in it that been solved since then.
We will need to enforce the fact that Eigen 3.4.1  should not be used.

nim65s:
I'm a bit lost on the eigen 3.4.1 issue. Could you share a link, or copy/paste an error you get, or explain a bit what is this issue ? 

On another topic, when you say "We will need to enforce the fact that Eigen 3.4.1 should not be used.", what do you mean, if https://github.com/humanoid-path-planner/hpp-fcl/commit/b2673f58bbdc2f9c0adc55c25b5a9b3b94e701f4 is not doing that ?

jcarpent:
> I'm a bit lost on the eigen 3.4.1 issue. Could you share a link, or copy/paste an error you get, or explain a bit what is this issue ?
> 
> On another topic, when you say "We will need to enforce the fact that Eigen 3.4.1 should not be used.", what do you mean, if [b2673f5](https://github.com/humanoid-path-planner/hpp-fcl/commit/b2673f58bbdc2f9c0adc55c25b5a9b3b94e701f4) is not doing that ?

Yes, sorry. It seems that the issue was only related to conda-forge, in fact. They have automated the process, and it has proceed the branch 3.4.1. See https://github.com/conda-forge/eigen-feedstock/issues/39 for further info.

PR #451, opened 2023-08-02
Convex and BVHModel: moving from raw to std::shared_ptr

lmontaut:
Related to #450.

Since we remove the concept of `own_storage` to use `std::shared_ptr` instead of raw pointers, this PR introduces a breaking change in the API, when creating `Convex` objects from scratch:
```cpp
// previously:
Vec3f* pts = new Vec3f[n_vertices];
Triangle* tris = new Vec3f[n_tris];
bool own_storage = true;
// fill pts and tris...
// then call the `Convex` constructor:
Convex<Triangle> cvx(own_storage, points, n_vertices, tris, n_tris);

// now:
std::shared_ptr<Vec3f> pts(new Vec3f[n_vertices]);
std::shared_ptr<Triangle> tris(new Vec3f[n_tris]);
// fill pts and tris...
// then call the `Convex` constructor:
Convex<Triangle> cvx = convexHull(points, n_vertices, tris, n_tris);
```

### DIFF
--- a/.github/workflows/ros_ci.yml
+++ b/.github/workflows/ros_ci.yml
@@ -13,7 +13,7 @@ jobs:
         env:
           - {ROS_DISTRO: melodic, PRERELEASE: false}
           - {ROS_DISTRO: noetic}
-          - {ROS_DISTRO: foxy}
+          - {ROS_DISTRO: iron}
           - {ROS_DISTRO: rolling}
     env:
       # CCACHE_DIR: /github/home/.ccache # Enable ccache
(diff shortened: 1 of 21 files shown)


nim65s (approved):
Thanks ! I only did a light review, but I think if the CI pass with this, this is totally OK :)

lmontaut:
Thanks @nim65s ! :) I also just added 2 commits which finalized the move from raw to shared ptrs in `BVHModel`.

The ROS CI seems to fails but I am not sure I understand why, it seems to be stuck when trying to generate the prerelease script, which does not seem related to the modifications of this PR(?):
```
RuntimeError: Unknown release: 'foxy'. Valid release names are: 'humble', 'iron', 'rolling'
```
I am not familiar at all with ROS, am I missing something?

PR #457, opened 2023-10-05
[ConvexBase] Fix neighbor copy in copy constructor.

lmontaut:
Previously, the pointers in `neighbors` were not pointing to the data stored in `nneighbors_` but to `other.nneighbors_`.
If `other.nneighbors_` is destroyed then the neighbors point to deleted data. This PR fixes that.


lmontaut:
@jcarpent it seems the ci is failing because the hppfcl3x branch doesn't have the fix we did for the python documentation (#455).

PR #473, opened 2023-11-14
[broadphase/detail] explicitly include <cstdint> header in morton.h

ManifoldFR:
This fixes a compilation bug (missing types `uint32_t` and `uint64_t`) on my setup (Clang-15 from `apt` on Ubuntu, so not unusual), and in the abstract sense it's good to make sure we explicitly include the header which defines these [typedefs](https://en.cppreference.com/w/cpp/types/integer).


nim65s (approved):
Thanks !

PS: @olivier-stasse: now we have an example where this is required ;)

PR #481, opened 2023-11-22
Performance Patch

CaeruleusAqua:
Hi all, I am currently profiling my application and have found a bottleneck in hpp fcl.

The collisionGeometry() function returns a copy of a shared_ptr. Unfortunately a reference is not possible, because a const cast is necessary.

So I built a function that returns a raw pointer, which solves the problem.

Im using the recent Visual Studio Compiler.

Here is the Profiler output, before the change:
image

image

And here with it:

image

### DIFF
--- a/include/hpp/fcl/collision_object.h
+++ b/include/hpp/fcl/collision_object.h
@@ -305,6 +305,11 @@ class HPP_FCL_DLLAPI CollisionObject {
     return cgeom;
   }
 
+  /// @brief get geometry from the object instance, as raw pointer
+  const CollisionGeometry* collisionGeometryRaw() const {
+    return cgeom.get();
+  }
+
   /// @brief get geometry from the object instance
   const shared_ptr<CollisionGeometry>& collisionGeometry() { return cgeom; }
 
--- a/include/hpp/fcl/collision_object.h
+++ b/include/hpp/fcl/collision_object.h
@@ -306,9 +306,7 @@ class HPP_FCL_DLLAPI CollisionObject {
   }
 
   /// @brief get geometry from the object instance, as raw pointer
-  const CollisionGeometry* collisionGeometryRaw() const {
-    return cgeom.get();
-  }
+  const CollisionGeometry* collisionGeometryRaw() const { return cgeom.get(); }
 
   /// @brief get geometry from the object instance
   const shared_ptr<CollisionGeometry>& collisionGeometry() { return cgeom; }
--- a/src/collision.cpp
+++ b/src/collision.cpp
@@ -62,8 +62,8 @@ void CollisionResult::swapObjects() {
 
 std::size_t collide(const CollisionObject* o1, const CollisionObject* o2,
                     const CollisionRequest& request, CollisionResult& result) {
-  return collide(o1->collisionGeometry().get(), o1->getTransform(),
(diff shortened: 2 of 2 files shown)


jcarpent:
Thanks @CaeruleusAqua for reporting this issue. Could you please provide a reproducing example to try on our side?
Best,
Justin

CaeruleusAqua:
Unfortunately, I cannot provide a simple example with fcl.

I have used the collision test as a benchmark (with a loop of 2000 collision check runs, in test/collision.cpp)
I get an average improvement of about 3%.
Certainly also depends on the mesh...

In my application, however, the difference was significantly greater (up to 20%).
However, it is also heavily multithreaded, and I think that's where the problem lies, because shared_ptr uses an atomic in the reference counter, which becomes slower as the number of threads increases, depending on the implementation.
REF: 
https://ddanilov.me/shared-ptr-is-evil/

The following example shows the general problem with the const cast in C++ in general:

```
#include <iostream>
#include <memory>
#include <vector>
#include <chrono>

class ComplexObject {
public:
    ComplexObject() : data(100) {
    }

    std::vector<int> data;

};

class Test {
public:
    std::shared_ptr<ComplexObject> cgeom;

    Test() {
        cgeom = std::make_shared<ComplexObject>();
    }

    const std::shared_ptr<ComplexObject> &collisionGeometry() {
        return cgeom;
    }

    const std::shared_ptr<const ComplexObject> collisionGeometryConst() const {
        return cgeom;
    }

    const ComplexObject *collisionGeometryRaw() const {
        return cgeom.get();
    }
};


int main() {
    const ComplexObject *a;
    Test obj;

    size_t counter = 10000000;

    {
        auto start_time = std::chrono::high_resolution_clock::now();
        for (int i = 0; i < counter; ++i) {
            a = obj.collisionGeometry().get();
            (void) a;
        }
        auto end_time = std::chrono::high_resolution_clock::now();
        auto time = end_time - start_time;
        std::cout << time / std::chrono::microseconds(1) << "ms to run. shared_ptr\n";
    }

    {
        auto start_time = std::chrono::high_resolution_clock::now();
        for (int i = 0; i < counter; ++i) {
            a = obj.collisionGeometryConst().get();
            (void) a;
        }
        auto end_time = std::chrono::high_resolution_clock::now();
        auto time = end_time - start_time;
        std::cout << time / std::chrono::microseconds(1) << "us to run. shared_ptr const\n";
    }

    {
        auto start_time = std::chrono::high_resolution_clock::now();
        for (int i = 0; i < counter; ++i) {
            a = obj.collisionGeometryRaw();
            (void) a;
        }
        auto end_time = std::chrono::high_resolution_clock::now();
        auto time = end_time - start_time;
        std::cout << time / std::chrono::microseconds(1) << "us to run. raw ptr const\n";
    }

}
```

Output:

```
0ms to run. shared_ptr
77509us to run. shared_ptr const
0us to run. raw ptr const
```

jcarpent:
Dear @CaeruleusAqua,
We appreciate your time and effort in finding this significant issue.
@lmontaut has provided a complete fix in #482, which is based on hppfcl-3x, the next major release that we will publish soon.
I will close this PR as it is redundant with this one.

Thank you again for your contribution :-)

PR #483, opened 2023-11-25
Import #482 and fix Python stubs testing on Windows

jcarpent:


jcarpent:
Related to #481 and #482.

PR #495, opened 2023-12-05
make CachedMeshLoader check file last modification time.

jmirabel:
This makes loading less error prone.


jcarpent:
@jmirabel The current implementation seems to be for Unix systems.
Maybe, we need to rely on boost directly to operate on all the supported systems.

jcarpent:
@jmirabel You should also update changelog file to account for your modifications ;)

jmirabel:
> @jmirabel The current implementation seems to be for Unix systems. Maybe, we need to rely on boost directly to operate on all the supported systems.

Here you go.

jcarpent (dismissed):

PR #525, opened 2024-01-29
Add unit tests for Box-Plane and Box-Halfspace collisions

Kotochleb:
As in https://github.com/humanoid-path-planner/hpp-fcl/pull/520 I recreated the unit test for the Box-Plane collision on this branch. I also redid the same tests for the Box-Halfspace and basically cleaned the up.

Currently, this PR is a draft since it some of the tests are failing. In both cases (Plane and Halfspace) the accuracy of position of the contact point in the new implementation got worse to the point where error is of magnitude greater than 1e-3. This only fails if Box is not orthogonal to the Plane/Halfspace. This is nicely visible in the tests for Box-Plane, as those are the only failing tests. As a side note, those tests are passing with the old implementation (at least for the Box-Plane)

In the case of the Box-Halfspace there are two issues. When the Box is intersecting with the Halfspace, the nearest point for the Box is the one farthest away from the surface of the Halfspace. In case the Box is fully "submerged" in the Halfspace the nearest point jumps to the center of the Box.

The second problem is that in the same test, when the box is standing at its edge, Box-Plane will return projection of the COG onto the plane, being the center of the edge, while Box-Halfspace will return the vertex of the Box which actually can be seen in screenshots in the previous https://github.com/humanoid-path-planner/hpp-fcl/pull/520.


jcarpent (approved):

jcarpent:
@lmontaut Could you have a look and provide an answer to @Kotochleb?

PR #527, opened 2024-02-05
Hide qhullcpp symbol

jorisv:
# Problem 1

qhullcpp is a static library that had been build with default symbol visibility strategies : all is visible.
When linking against this library, hpp-fcl get all libqhull symbols and also show them as visible.

If a thirdparty library is linking against hpp-fcl and qhullcpp, then a strange symbol mix-up can occur.
I have been able to reproduce it with pinocchio 3.

This can lead to binary conflicts, since there is no guarantee that pinocchio 3 and hpp-fcl are using the same version of qhullcpp with the same ABI (same compiler flags).

## Solution

I advise to hide all qhullcpp symbols when linking against it.
I only know how to do it and test it on Linux. We should find a way to fix it on MacOS and Windows if the same issue occur.

# Problem 2

hpp-fcl embed qhullcpp source code. This source code is working with some old qhull_r versions. I have not be able to build it with qhull_r 2020.2.

## Solution

Maybe we can get rid of it ? Or at least document which version of qhull_r is needed.

If we keep it, then we should find a way to not export the qhullcpp symbol either (probably by creating a static library).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -214,6 +214,10 @@ if(HPP_FCL_HAS_QHULL)
   target_compile_definitions(${LIBRARY_NAME} PRIVATE -DHPP_FCL_HAS_QHULL)
   if (HPP_FCL_USE_SYSTEM_QHULL)
     target_link_libraries(${LIBRARY_NAME} PRIVATE Qhull::qhull_r Qhull::qhullcpp)
+    if(UNIX)
+      set_target_properties(
+        ${LIBRARY_NAME} PROPERTIES LINK_FLAGS "-Wl,--exclude-libs,libqhullcpp")
+    endif()
   else()
     target_include_directories(${LIBRARY_NAME} SYSTEM PRIVATE
       ${Qhull_r_INCLUDE_DIR} ${Qhullcpp_PREFIX})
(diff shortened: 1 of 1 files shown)


jorisv:
It's maybe possible to hide symbol in macos with [-hidden-lx](https://stackoverflow.com/questions/58580964/workaround-exclude-libs-not-available-on-clang-for-os-x) linker flag.

On Windows, I think that symbol are hidden by default (no dllexport).

jorisv:
Hello @jmirabel,

We speak with @nim65s yesterday about this issue.
You left the following explanation on why we compile qhullcpp :

> TODO We compile qhullcpp because it is not provided in the binary package while the other parts of Qhull are released.

Do you remember which package system doesn't provide qhullcpp ?

I quickly check on ubuntu with docker:

|ubuntu|version|qhull|qhull_r|qhullcpp|
|---------|----|------|---------|----------|
|20.04|2015.2-4|x|x||
|22.04|2020.2-4|x|x|x|
|24.04|2020.2-6|x|x|x|

Do you have any other packaging system in mind ?

jmirabel:
This is an old problem and I may not remember everything perfectly. It's older than 20.04 for sure.

The issue was that the package found on the official repos did not contain all the needed libraries to link to it.

I'm strongly in favor of removing the internal qhull package if using the libs provided on the official repos works. From your table, it seems that it won't work on 20.04.

What is the minimal ubuntu version that is to be supported?

nim65s:
ubuntu LTS support is 5y, so 20.04 + 5y -> we can drop that on May 1st :)

jorisv:
Ok, we can remove the support on may 1st then.

Since Ubuntu provide shared library, I think we don't need to use custom instructions to hide static library symbols. So this PR will be probably simplified.
I will make some tests to be sure using the shared library doesn't create any issue.

jorisv:
Done in #704

Issue #380, opened 2023-03-13
Interpretation of Assertions 


cologne86:
Hi,
Using this library with `NDEBUG` set i am experiencing crashes caused by `assert` calls.

I am getting assertions using two `hpp::fcl::BVHModel<hpp::fcl::OBBRSS>` objects as input
calling `hpp::fcl::collide`. Depending on transformations those would sometimes cause one of the following asserts:

https://github.com/humanoid-path-planner/hpp-fcl/blob/893d57c6e08434fd482a656977c5376cb360e6b4/src/narrowphase/gjk.cpp#L331
https://github.com/humanoid-path-planner/hpp-fcl/blob/893d57c6e08434fd482a656977c5376cb360e6b4/src/narrowphase/gjk.cpp#L558
https://github.com/humanoid-path-planner/hpp-fcl/blob/893d57c6e08434fd482a656977c5376cb360e6b4/src/narrowphase/gjk.cpp#L1264

However, when looking at the surrounding code here:

```
       if (la > lb) {
          assert(false);
          w0 = b0;
          w1 = b1;
```

it seems that these assertions are not explicitly error states,but the algorithm does handle the input in that case. Constraints in other cases also seem to me quite strict.

When building the software with `NDEBUG`, results still look valid and i have not experienced any crashes.
The question is:
Should the assertions never be triggered with a valid input, or is it OK to go on without debug information?

lmontaut:
Hi @cologne86,
Could you please provide an example of code to replicate and fix the bugs?

The asserts correspond to degenerate cases that can happen with a valid input.
> https://github.com/humanoid-path-planner/hpp-fcl/blob/893d57c6e08434fd482a656977c5376cb360e6b4/src/narrowphase/gjk.cpp#L331

This assert failing is due to GJK currently using a relative convergence criterion making the value `1e-6` arbitrary. This will be fixed in hppfcl3x notably by switching to an absolute convergence criterion.

> https://github.com/humanoid-path-planner/hpp-fcl/blob/893d57c6e08434fd482a656977c5376cb360e6b4/src/narrowphase/gjk.cpp#L558
 
This assert signals a rare behavior when computing nearest points between two shapes, related to how GJK constructs the simplex leading to the computation of nearest points. The code after the assert is the fallback behavior of hppfcl. The computed nearest points remain valid however.

[...]

PR #531, opened 2024-02-08
Setting default parameters for the narrowphase

lmontaut:
This PR is related to #380 where users are sometimes experiencing assertion errors in Debug mode.
This PR is related to one of these assertion triggers:
- When the user sets a tolerance for GJK and/or EPA that is too low, the support function may work with support directions that are almost zero, which is not suitable for GJK and EPA. In hppfcl, we have hard-coded assertion checks, and these checks make the assumption that GJK's tolerance is not below `1e-6`. This PR introduces default values for GJK and EPA and places a warning in Debug mode only for users going below this tolerance.
Side note: when working with the default GJK and EPA (default `1e-6` tolerance, default convergence criterion), it is precise up to the micro-meter, which is an **extremely** low tolerance. In practice, working with `1e-4` or even `1e-3` should be more than enough.

Sorry for the long coming text.
I will use this PR to evoke the 2 main other assertions that sometimes trigger:
- When shapes of a collision pair are extremely far apart or have extremely different dimensions (i.e., a molecule vs. a plane), the projection step in GJK may be limited due to floating point error, and assertions will be triggered. We might want to implement a "pre-conditioning" system in the future. When working with "molecule vs. plane" dimensions, it would warn the user that hppfcl is not made to work with such different dimensions. When working with "molecule vs. molecule" dimensions, it would simply scale the problem "to the meter." This pre-conditioner would simply check the dimension of the AABB of each shape.
- When shapes of a collision pair are in "touching contact", i.e they are colliding but have a 0 penetration depth. In such a case, EPA tries to run, fails, and an assert triggers. Although this situation in "rare" in theory, it actually occurs quite often when using hppfcl for physics simulation. I am actively working on fixing this issue, so hopefully this assert will soon disappear.

### DIFF
--- a/include/hpp/fcl/narrowphase/narrowphase_defaults.h
+++ b/include/hpp/fcl/narrowphase/narrowphase_defaults.h
@@ -1,0 +1,58 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2024, INRIA
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of Open Source Robotics Foundation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/// This file defines different macros used to characterize the default behavior
+/// of the narrowphase algorithms GJK and EPA.
+
+#ifndef HPP_FCL_NARROWPHASE_DEFAULTS
+#define HPP_FCL_NARROWPHASE_DEFAULTS
+
+/// GJK
+#define GJK_DEFAULT_MAX_ITERATIONS 128
+#define GJK_DEFAULT_TOLERANCE 1e-6
+/// Note: if the considered shapes are on the order of the meter, and the
+/// convergence criterion of GJK is the default VDB criterion,
+/// setting a tolerance of 1e-6 on the GJK algorithm makes it precise up to
+/// the micro-meter.
+/// The same is true for EPA.
+#define GJK_MINIMUM_TOLERANCE 1e-6
+
+/// EPA
+#define EPA_DEFAULT_MAX_ITERATIONS 255
+#define EPA_DEFAULT_TOLERANCE 1e-6
+#define EPA_MINIMUM_TOLERANCE 1e-6
+#define EPA_DEFAULT_MAX_FACES 128
+#define EPA_DEFAULT_MAX_VERTICES 64
+
+#endif  // HPP_FCL_NARROWPHASE_DEFAULTS
--- a/include/hpp/fcl/narrowphase/narrowphase_defaults.h
+++ b/include/hpp/fcl/narrowphase/narrowphase_defaults.h
@@ -38,21 +38,29 @@
 #ifndef HPP_FCL_NARROWPHASE_DEFAULTS
 #define HPP_FCL_NARROWPHASE_DEFAULTS
 
+#include <hpp/fcl/data_types.h>
+
+namespace hpp {
+namespace fcl {
+
 /// GJK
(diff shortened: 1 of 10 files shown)


lmontaut:
This PR also fixes additional assertion errors in GJK and BVH thanks to tests provided by #382

jcarpent (approved):

test/collision_node_asserts.cpp
```diff
@@ -0,0 +1,57 @@
+#define BOOST_TEST_MODULE FCL_COLLISION_NODE_ASSERT
+
```
jcarpent, line 2:
```suggestion

#define _USE_MATH_DEFINES
```

jcarpent (approved):

include/hpp/fcl/narrowphase/narrowphase_defaults.h
```diff
@@ -0,0 +1,58 @@
...
+/// the micro-meter.
+/// The same is true for EPA.
+#define GJK_MINIMUM_TOLERANCE 1e-6
+
+/// EPA
+#define EPA_DEFAULT_MAX_ITERATIONS 255
+#define EPA_DEFAULT_TOLERANCE 1e-6
+#define EPA_MINIMUM_TOLERANCE 1e-6
+#define EPA_DEFAULT_MAX_FACES 128
+#define EPA_DEFAULT_MAX_VERTICES 64
```
jorisv, line 56: Instead of define, use global constant and put in a namespace.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es31-dont-use-macros-for-constants-or-functions

src/collision_node.cpp
```diff
@@ -53,13 +53,16 @@ void collide(CollisionTraversalNodeBase* node, const CollisionRequest& request,
...
+    HPP_FCL_UNUSED_VARIABLE(dummy_precision);
     if (!std::isnan(sqrDistLowerBound)) {
       if (sqrDistLowerBound == 0) {
-        assert(result.distance_lower_bound <= 0);
+        assert(result.distance_lower_bound <= dummy_precision);
       } else {
         assert(result.distance_lower_bound * result.distance_lower_bound -
                    sqrDistLowerBound <
-               1e-8);
+               dummy_precision);
```
jorisv, line 65: This code is checking if the result is correct.

1. Are you sure throwing an exception is not a better behavior ?
2. I would put this code in a separated static function (for the symbol visibility) to split the useful code from the error checking.

src/narrowphase/gjk.cpp
```diff
@@ -1094,7 +1095,10 @@ bool GJK::projectTetrahedraOrigin(const Simplex& current, Simplex& next) {
   const Vec3f a_cross_b = A.cross(B);
   const Vec3f a_cross_c = A.cross(C);
 
-  const FCL_REAL dummy_precision = Eigen::NumTraits<FCL_REAL>::epsilon();
+  // dummy_precision is 1e-14 if FCL_REAL is double
+  //                    1e-5  if FCL_REAL is float
+  static const FCL_REAL dummy_precision(
+      100 * std::numeric_limits<FCL_REAL>::epsilon());
```
jorisv, line 1101: I think you can remove the static keyword here.

test/collision_node_asserts.cpp
```diff
@@ -0,0 +1,58 @@
...
+#define _USE_MATH_DEFINES
+#include <cmath>
```
jorisv, line 4: Instead of using _USE_MATH_DEFINES you can use `boost::math::constants::pi<Real>()` from `#include <boost/math/constants/constants.hpp>`

test/gjk_asserts.cpp
```diff
@@ -0,0 +1,90 @@
+#define BOOST_TEST_MODULE FCL_GJK_ASSERTS
+
+#define _USE_MATH_DEFINES
+#include <cmath>
```
jorisv, line 4: Same than previous issue

test/collision_node_asserts.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <boost/test/included/unit_test.hpp>
+#include <hpp/fcl/BVH/BVH_model.h>
+#include <hpp/fcl/collision.h>
+
+using namespace hpp::fcl;
+
+inline double DegToRad(const double& deg) {
```
jorisv, line 11:
```suggestion
double DegToRad(const double& deg) {
```

test/gjk_asserts.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <boost/test/included/unit_test.hpp>
+#include <hpp/fcl/BVH/BVH_model.h>
+#include <hpp/fcl/collision.h>
+
+using namespace hpp::fcl;
+
+inline double DegToRad(const double& deg) {
```
jorisv, line 11:
```suggestion
double DegToRad(const double& deg) {
```

test/gjk_asserts.cpp
```diff
@@ -0,0 +1,90 @@
...
+  static double degToRad = M_PI / 180.;
+  return deg * degToRad;
+}
+std::vector<Vec3f> dirs{Vec3f::UnitZ(),  -Vec3f::UnitZ(), Vec3f::UnitY(),
+                        -Vec3f::UnitY(), Vec3f::UnitX(),  -Vec3f::UnitX()};
+
+inline void CreateSphereMesh(BVHModel<OBBRSS>& model, const double& radius) {
```
jorisv, line 18:
```suggestion
void CreateSphereMesh(BVHModel<OBBRSS>& model, const double& radius) {
```

test/gjk_asserts.cpp
```diff
@@ -0,0 +1,90 @@
...
+  size_t polarSteps{32};
+  size_t azimuthSteps{32};
+
+  const float PI = static_cast<float>(M_PI);
```
jorisv, line 22: boost::math::constants::pi

test/gjk_asserts.cpp
```diff
@@ -0,0 +1,90 @@
...
+            Eigen::AngleAxis<double>(DegToRad(j), Vec3f::UnitY()));
+        for (const Vec3f& dir : dirs) {
+          sphere2Tf.setTranslation(dir);
+          CollisionResult result;
+
+          BOOST_CHECK_NO_THROW(compute(sphere2Tf, sphere1Tf, request, result));
+        }
+      }
+    }
+  }
```
jorisv, line 89: I don't get the purpose of this test. It test if the assertion in collision_node are not trigger ?

I find it weird to test an assertion (the test is only valid in DEBUG mode). Also, I would test that te assertion IS thrown and not that there is no assertion (you don't validate the error case).

jorisv (changes requested):

test/gjk_asserts.cpp
lmontaut: I agree with you @jorisv. The thing is: without this test, I would not have been able to debug the assertion in GJK. Now that the assertion is fixed, this test is basically useless and will always pass...
Also, we can't really design a test that **does** trigger the GJK assertion because the point is that we **don't** want the GJK assertion to trigger.

What do you think is good practice? Should we remove this test?

src/collision_node.cpp
lmontaut: I think you're right, this should probably throw an exception.

Contrary to the assertions in GJK for which GJK has a fallback behavior and will still work fine, here if the asserts are triggered it means the entire collide call is wrong.
So throwing an exception is probably better.

lmontaut: Got update from @jcarpent.
This is deep in the code and the throw will slow things down. I will keep the function but replace the throws by asserts.

jorisv (approved):

PR #533, opened 2024-02-12
Add option to turn asserts into exceptions and fix EPA related assert

lmontaut:


CMakeLists.txt
```diff
@@ -130,6 +130,8 @@ else()
   CHECK_MINIMAL_CXX_STANDARD(14 ENFORCE)
 endif()
 
+option(HPP_FCL_TURN_ASSERT_INTO_EXCEPTION "turn some critical HPP-FCL asserts to exception." FALSE)
```
jcarpent, line 133: Could you please move this line close to the other options for readibility?

jcarpent (changes requested):

jorisv (changes requested):
I see other assert in src and include. Are you sure none of them should be turn into exception (velocity issue, assertion that really should never happen) ?

jcarpent (approved):

PR #535, opened 2024-02-15
Improve EPA

lmontaut:
This PR is quite big, apologies for that in advance.
It does multiple things:
- It removes the successive mallocs in EPA when calling `GJKSolver`'s methods successively.
- To do so, it was necessary to refactor `GJKSolver`. Refactoring it removes copy-pastes between the different methods of the class, handling **all possible statuses** of GJK and EPA once they are called, adding documentation, adding logging to warn the user about problematic cases (only on debug mode or by activating the `HPP_FCL_ENABLE_LOGGING` preprocessor option).
- The fields `enable_contact` of `CollisionRequest` and `enable_signed_distance` of `DistanceRequest` now allow the user to make a trade-off between performance vs. contact information (sometimes we only want to know if two objects collide and only want to compute their distance when they are disjoint).
- It fixes the convergence criterion of EPA. The algo should be more robust now.

Because all of the statuses of GJK/EPA are now clear in the code, the truly degenerate cases of EPA (non-convex, invalid hull etc.) now trigger an assertion in debug mode. It turns out that some of the tests that we had do trigger these assertions (like `test/geometric_shapes.cpp` on cones or cylinders), which is good because it allows to understand the behavior of EPA. These tests are fixed now (EPA's tolerance cannot be too low on strictly-convex shapes, simply because it has terrible performance on them, which is expected from the algorithm).


include/hpp/fcl/narrowphase/gjk.h
```diff
@@ -150,13 +150,25 @@ struct HPP_FCL_DLLAPI GJK {
...
+  /// class.
   struct HPP_FCL_DLLAPI Simplex {
     /// @brief simplex vertex
     SimplexV* vertex[4];
     /// @brief size of simplex (number of vertices)
     vertex_id_t rank;
 
     Simplex() {}
+
+    inline void reset() {
```
jorisv, line 168:
```suggestion
    void reset() {
```

Methods defined in a class/struct are inline by default.

src/narrowphase/gjk.cpp
```diff
@@ -525,13 +525,17 @@ void MinkowskiDiff::set(const ShapeBase* shape0, const ShapeBase* shape1) {
...
   gjk_variant = GJKVariant::DefaultGJK;
   convergence_criterion = GJKConvergenceCriterion::VDB;
   convergence_criterion_type = GJKConvergenceCriterionType::Relative;
+  reset();
+}
+
+void GJK::reset() {
+  status = DidNotRun;
+  nfree = 0;
+  simplex = NULL;
```
jorisv, line 538:
```suggestion
  simplex = nullptr;
```

include/hpp/fcl/narrowphase/gjk.h
```diff
@@ -379,15 +398,26 @@ struct HPP_FCL_DLLAPI EPA {
...
+  size_t iterations;
+
   Status status;
   GJK::Simplex result;
   Vec3f normal;
   FCL_REAL depth;
+
+ private:
   SimplexV* sv_store;
   SimplexF* fc_store;
```
jorisv, line 416: Why using raw ptr and managing memory manually ?

I think you can use safely a std::vector here.

include/hpp/fcl/narrowphase/gjk.h
```diff
@@ -397,12 +427,32 @@ struct HPP_FCL_DLLAPI EPA {
...
+  /// @brief Copy constructor of EPA.
+  /// Mostly needed for the copy constructor of `GJKSolver`.
+  EPA(const EPA& other)
+      : max_face_num(other.max_face_num),
+        max_vertex_num(other.max_vertex_num),
+        max_iterations(other.max_iterations),
+        tolerance(other.tolerance) {
+    initialize();
+    memcpy(sv_store, other.sv_store, max_vertex_num * sizeof(SimplexV));
+    memcpy(fc_store, other.fc_store, max_face_num * sizeof(SimplexF));
```
jorisv, line 439: If you manage to turn sv_store and fc_store into a std::vector, you can use the default copy constructor.

src/CMakeLists.txt
```diff
@@ -202,6 +202,12 @@ TARGET_LINK_LIBRARIES(${LIBRARY_NAME}
   Boost::filesystem
 )
 
+if (HPP_FCL_ENABLE_LOGGING)
+  TARGET_LINK_LIBRARIES(${LIBRARY_NAME} PUBLIC Boost::log)
+  # The compile flag `BOOST_LOG_DYN_LINK` is required here.
+  target_compile_definitions(${LIBRARY_NAME} PUBLIC -DHPP_FCL_ENABLE_LOGGING -DBOOST_LOG_DYN_LINK)
```
jorisv, line 208:
```suggestion
  target_compile_definitions(${LIBRARY_NAME} PUBLIC HPP_FCL_ENABLE_LOGGING BOOST_LOG_DYN_LINK)
```

include/hpp/fcl/narrowphase/gjk.h
jorisv: If for some reason you can't use std::vector, try to use std::unique_ptr to add more safety.

jorisv (dismissed):

include/hpp/fcl/narrowphase/gjk.h
lmontaut: Using std::vector now. Thanks @jorisv !

CMakeLists.txt
```diff
@@ -116,8 +117,14 @@ SET_BOOST_DEFAULT_OPTIONS()
 EXPORT_BOOST_DEFAULT_OPTIONS()
 IF(WIN32)
   ADD_PROJECT_DEPENDENCY(Boost REQUIRED COMPONENTS chrono thread date_time serialization filesystem)
+  if (HPP_FCL_ENABLE_LOGGING)
+    ADD_PROJECT_DEPENDENCY(Boost REQUIRED COMPONENTS log)
+  endif()
 ELSE(WIN32)
   ADD_PROJECT_DEPENDENCY(Boost REQUIRED chrono serialization filesystem)
+  if (HPP_FCL_ENABLE_LOGGING)
+    ADD_PROJECT_DEPENDENCY(Boost REQUIRED log)
```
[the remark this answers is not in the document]
lmontaut: Ok I will merge the two if statements and keep all the boost dependencies. Let see if it builds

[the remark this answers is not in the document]
lmontaut: Yep, seems to build  just fine on windows. Thanks!

PR #536, opened 2024-02-19
Fix Heigth Field collision check

jcarpent:


include/hpp/fcl/internal/traversal_node_hfield_shape.h
```diff
@@ -190,62 +241,166 @@ inline Vec3f projectTriangle(const Vec3f& pointA, const Vec3f& pointB,
...
+  const Vec3f pointB = points[triangle[1]];
+  const Vec3f pointC = points[triangle[2]];
+
+  const Vec3f contact_point_projected =
+      projectTriangle(pointA, pointB, pointC, contact_point);
+
+  return contact_point_projected;
+}
+
+inline FCL_REAL distanceContactPoinToTriangle(
```
lmontaut, line 282: Typo in function name -> `distanceContactPointToTriangle`

include/hpp/fcl/internal/traversal_node_hfield_shape.h
```diff
@@ -117,31 +134,65 @@ void buildConvexTriangles(const HFNode<BV>& node, const HeightField<BV>& model,
...
+    (*triangles)[1].set(3, 4, 5);  // top
+    (*triangles)[2].set(0, 1, 3);  // West 1
+    (*triangles)[3].set(3, 1, 4);  // West 2
+    (*triangles)[4].set(1, 2, 5);  // South-East 1
+    (*triangles)[5].set(1, 5, 4);  // South-East 1
+    (*triangles)[6].set(0, 5, 2);  // North 1
+    (*triangles)[7].set(5, 0, 3);  // North 2
 
     convex1.set(pts,  // points
                 6,    // num points
```
lmontaut, line 198: Can't we store `convex1` and `convex2` in the HeightField? This would avoid some allocations and uselessly calling `fillNeighbors` of `convex1` and `convex2` (neighbors will never be used in this case).

lmontaut (approved):

include/hpp/fcl/internal/traversal_node_hfield_shape.h
jcarpent: It looks to be indeed a nice improvement. We might thus plan it in following PR.
Thanks @lmontaut for this nice suggestion.

PR #556, opened 2024-03-21
Misc gjk epa

lmontaut:
Multiple little GJK/EPA modifications:
- [API change] Renamed default convergence criterion from `VDB` to `Default`
- Renamed exit status of GJK (no API change, only used internally)
- Removed useless function I previously introduced in EPA
- If EPA encounters badly conditioned problems, do not return nans. Often, the `Degenerate`, `NonConvex` and `InvalidHull` cases, the estimate of the normal and penetration depth is not the best but it's better than returning nans.


jcarpent (approved):
I made a quick pass and the change seems fine to me.
Thanks @lmontaut for keeping this track of cleaning and enhancing GJK/EPA algos.

PR #558, opened 2024-03-29
Adding swept-sphere radius + removing dead code

lmontaut:
Ok, it's a big PR, haven't finished it yet but it's almost done.
Before I rant, note that there are absolutely no changes to HPP-FCL API, it's only internal stuff.

It started with a simple premise: adding swept-sphere radius for all shapes of HPP-FCL.
However, it got a bit out of hand.

When making simple modifications to the library, I would always encounter compilation errors which were coming from dead code.
This is very frustrating to deal with because 90% of the time is spent fixing code that is not used and not supported.
Most of the problems come from a single culprit: the `GJKSolver`. This solver had specializations for specific pairs of shapes.
However, these specializations were already dealt with in the `ShapeShapeCollide/Distance` internal functions. So this means having to keep track of multiple specializations. On top of that, most of these specializations were located in a `details.h` file which contained about 2000 LOC of dead code. Also, many functions had conflicting convention on their return value.

The solution: I moved all the specialized algorithms to the `ShapeShapeCollide/Distance` internal functions and skinned the `GJKSolver`. Now `GJKSolver`, as its name indicates, is **only used** to run GJK/EPA.
This also has the benefit of being able to test the specialized algorithms against GJK/EPA.

The overall philosophy is the following: no internal functions in HPP-FCL should directly call the `GJKSolver`.
All internal functions should use either `ShapeShapeCollide` or `ShapeShapeDistance`, which take care of either using a specialized algorithm or GJK/EPA.

TODOs regarding the original swept-sphere PR:
- [x] Finish swept-sphere for specialized functions (sphere-box, sphere-cylinder, plane-plane, halfspace-halfspace, plane-halfspace)
- [x] Rename the template parameter of support functions from `Inflation[...]` to `SweptSphere[...]`
- [x] Export template specializations of `MinkowskiDiff::set` and `getSupport`

Next PR:
- [ ] Take into account swept-sphere in AABB/OBB computation
- [ ] Add swept-sphere for BVHs, Hfields and Octrees. Also make sure Broadphase takes swept-sphere into account (it should thanks to `computeLocalAABB`).


florent-lamiraux:
Hello @lmontaut 
By swept sphere radius, do you mean the Minkowski sum of a shape with a sphere ?

lmontaut:
> Hello @lmontaut By swept sphere radius, do you mean the Minkowski sum of a shape with a sphere ?

Hi @florent-lamiraux, yes!
Just to clarify, this was already a feature of the `MinkowskiDiff` in HPP-FCL, I simply extracted it to make it available for all shapes (other than spheres and capsules).
There is no need to take the swept-sphere radius into account in the iterations of GJK/EPA.
However, it is sometimes usefull to compute the support function of shapes while taking into account their swept-sphere radius, so I added the support for that.

florent-lamiraux:
Isn't it redundant with security margins ?

lmontaut:
> Isn't it redundant with security margins ?

Yes, it's a redundant information if only the distance between the shapes is considered.
But it does change the witness points (by translating them along the normal).
I agree it's a simple transformation but it would require all the algos using the GJKSolver to factor in whether or not to modify the witness points (certain algos like HeightField-Shape do some internal bin corrections which need multiple call to GJK/EPA).

In the end, I think swept-sphere radius and security margin are two complementary information:
- The security margin is aimed at a pair of collision. It's used to make sure two shapes don't get closer than a certain value.
- A swept-sphere radius is a property of a shape. Sometimes it's usefull to described a rounded box by its underlying box + swept sphere radius, and not have to manually correct the witness points/distance by using the security margin.

include/hpp/fcl/collision_data.h
```diff
@@ -127,11 +127,10 @@ struct HPP_FCL_DLLAPI Contact {
         b1(b1_),
         b2(b2_),
         normal(normal_),
+        nearest_points{pos_ - (depth_ * normal_ / 2),
+                       pos_ + (depth_ * normal_ / 2)},
         pos(pos_),
-        penetration_depth(depth_) {
-    nearest_points[0] = pos - 0.5 * depth_ * normal_;
-    nearest_points[1] = pos + 0.5 * depth_ * normal_;
-  }
```
jcarpent, line 134: Small comment: changing these lines was not requested and should not provide any benefit.

python/gjk.cc
```diff
@@ -67,23 +67,68 @@ void exposeGJK() {
...
+            bp::args("self", "dir", "hint", "inflate_support"),
+            "Support of shape1 of the Minkowski difference. Set "
+            "`inflate_support` to true to take into account the swept sphere "
+            "radius of the shape.")
+
+        .def("support", &MinkowskiDiff::support,
+             "Compute Supports of shape0 and shape1 of the Minkowski "
+             "difference.",
+             (arg("dir"), arg("supp0"), arg("supp1"), arg("hint")))
+
```
jcarpent, line 131: Why did you change the way to parse the doc (previously, it was done using the doxygen info)?

jcarpent (approved):
Thanks @lmontaut for this massive and very useful clean up.
It will help us to remove outdated and unsupported codes.

florent-lamiraux:
> > Isn't it redundant with security margins ?
> 
> Yes, it's a redundant information if only the distance between the shapes is considered. But it does change the witness points (by translating them along the normal). I agree it's a simple transformation but it would require all the algos using the GJKSolver to factor in whether or not to modify the witness points (certain algos like HeightField-Shape do some internal bin corrections which need multiple call to GJK/EPA).
> 
> In the end, I think swept-sphere radius and security margin are two complementary information:
> 
>     * The security margin is aimed at a pair of collision. It's used to make sure two shapes don't get closer than a certain value.
> 
>     * A swept-sphere radius is a property of a shape. Sometimes it's usefull to described a rounded box by its underlying box + swept sphere radius, and not have to manually correct the witness points/distance by using the security margin.

All right. I understand.
I second @jcarpent : Thank you very much for cleaning up unused code.

jcarpent:
> * Serialize `CollisionGeometry` with `getNodeType`. For now, when we serialize it, we don't serialize the child object

Serialization works normally for derived class. @lmontaut Have you encountered any issue?

include/hpp/fcl/collision_data.h
lmontaut: I agree, I should have at least put than in another commit.
I originally wrote these lines but found them ugly...

lmontaut:
> > * Serialize `CollisionGeometry` with `getNodeType`. For now, when we serialize it, we don't serialize the child object
> 
> Serialization works normally for derived class. @lmontaut Have you encountered any issue?

Yes, I can't serialize a `CollisionGeometry` to a text file for example, but maybe I am doing something wrong... Also, on the python side, I forgot to add the serializable visitor to the python object `CollisionGeometry`.

Again, maybe I am doing something wrong on the C++ side. I will delete this todo after investigation :)

python/gjk.cc
lmontaut: Apologies, I update that with the doxygen way

PR #562, opened 2024-04-02
[CI] Turn off ROS2 Rolling to avoid failure

wxmerkt:
Since Noble isn't released yet, there are some hickups. Turning off prerelease should work around it for now.


nim65s (approved):

nim65s:
Thanks !

PR #579, opened 2024-05-15
Fix potential with older version of Eigen

jcarpent:

PR #580, opened 2024-05-15
ci pip: no py 3.8 on macos arm64 for now

nim65s:
Hi,

github runners "macos-latest" are now arm64 machines, and the previous build system for darwin arm was not able to generate python 3.8 builds.

Let's deactivate that for now, until I get some time to publish builds for those.


jcarpent:
Thanks @nim65s for the fix. 
Should be merged after #579.

jcarpent (approved):

PR #596, opened 2024-06-24
Renamming the library to `Coal`

lmontaut:
This PR addresses #593 and aims at changing the name of the library from `hpp-fcl` to `coal`.

**For the end-user**, this renaming implies the following changes:
- In CMake: `find_package(hpp-fcl)` -> `find_package(coal)`.
- In C++: `#include <hpp/fcl/...>` -> `#include <coal/...>`.
- In Python: `import hppfcl` -> `import coal`.

**C++ users**
I have added a layer of compatibility such that the namespace `hpp::fcl::` is still available and points to `coal::` if the cmake option `COAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL` is activated.
This option is set to false by default to encourage the use of `coal::` instead of `hpp::fcl::`.

I also took the opportunity to change and clarify certain fundamental types used throughout the library:
- `FCL_REAL` -> `CoalScalar`.
- `Matrix/Vec[...]f` (i.e. `Vec3f`) -> `Matrix/Vec[...]s` (i.e. `Vec3s`).
- `Transform3f` -> `Transform3s`.
The idea behind this change is that multiple people have reported that they found the `f` at the end of these types confusing because they thought it referred to `float` when the library is `double` by default.
Like the namespace, I kept the old data types behind the `COAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL` if needed.

**Things left to do**:
- [x] update readme
- [x] github workflow
- [ ] github repo name
- [x] update changelog

Things to do after PR is merged: conda build, package.xml, ROS.


jcarpent:
Should wait for #598 to be merged.

wxmerkt:
For ROS, we may need to make a set of fresh releases to avoid breaking existing packages (and it'll take some time to get dialed in during which any updated downstream projects may fail). Ping me when it's merged and I can get started on the weekend(s)

wxmerkt:
@jcarpent One more thought - would you be open to us first fixing all the warnings before renaming so we can stop the email flood for "unstable" builds for the then no-longer-releasable hpp-fcl library?

jcarpent:
> @jcarpent One more thought - would you be open to us first fixing all the warnings before renaming so we can stop the email flood for "unstable" builds for the then no-longer-releasable hpp-fcl library?

Yes, sure. @lmontaut will integrate your change into the renamed lib.
Thanks a lot @wxmerkt for your nice contributions.

nim65s:
> * In CMake: `find_package(hpp-fcl)` -> `find_package(coal)`.
> * In C++: `#include <hpp/fcl/...>` -> `#include <coal/...>`.
> * In Python: `import hppfcl` -> `import coal`.

Those alone can require quite significant work. I would be in favor of providing an hpp-fcl CMake package, hpp/fcl include dir, and hppfcl python module which would raise some warning when used, but then redirect / symlink / alias to coal.

I think we should be able to least configure/build/test pinocchio in its current form without requiring any change.

jorisv:
Hello @nim65s,

I agree that providing a full CMake/Python/C++ compatibility is a nice option.

I also think this option should not be activated in coal package (conda, nix, apt, ...). Since coal don't have the same name than hpp-fcl, we will be allowed by many packaging system to install coal and hpp-fcl at the same time. In this case, the compatibility files will conflict with the hpp-fcl files and create some hard to debug issues.

One option can be to release a hpp-fcl v3 package that depend of coal and provide the compatibility files.
Another option is to forbid the installation of hpp-fcl and coal in the same time, but I don't know if all package manager provide this option.

nim65s:
Yes, from a package manager point of view, coal should replace hpp-fcl. In archlinux terms, this is `conflicts=('hpp-fcl')` + `provides=('hpp-fcl')`. In python terms, this is `Obsoletes-Dist: hpp-fcl` for coal, and `Provides-Dist: coal` for hpp-fcl.
For monorepo package managers like robotpkg or nix, we don't care.

> One option can be to release a hpp-fcl v3 package that depend of coal and provide the compatibility files.

I already implemented that, and I'm testing this solution in various scenarii.

nim65s:
ref. https://github.com/lmontaut/hpp-fcl/pull/1

jorisv (approved):

nim65s:
I'd prefer  if we could decouple the workspace change from the rename change. This PR is already impossible to review.

src/CMakeLists.txt
```diff
@@ -187,6 +187,7 @@ add_library(${LIBRARY_NAME}
   ${PROJECT_HEADERS_FULL_PATH}
   ${${LIBRARY_NAME}_SOURCES}
   )
+add_library(hpp-fcl::hpp-fcl ALIAS ${LIBRARY_NAME})
```
nim65s: I would expect this to break with COAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL, because we can't define twice the same target

python/CMakeLists.txt
```diff
@@ -148,10 +148,15 @@ TARGET_LINK_LIBRARIES(${PYTHON_LIB_NAME} PUBLIC
 SET_TARGET_PROPERTIES(${PYTHON_LIB_NAME} PROPERTIES
   PREFIX ""
   SUFFIX "${PYTHON_EXT_SUFFIX}"
+<<<<<<< HEAD
```
nim65s: this probably should not be in git history

CMakeLists.txt
```diff
@@ -123,6 +123,7 @@ CMAKE_DEPENDENT_OPTION(GENERATE_PYTHON_STUBS "Generate the Python stubs associat
 ADD_PROJECT_DEPENDENCY(Eigen3 REQUIRED PKG_CONFIG_REQUIRES "eigen3 >= 3.0.0")
 
 if(BUILD_PYTHON_INTERFACE)
+  FINDPYTHON(REQUIRED)
```
nim65s, line 131: What is the goal of this ?
This is already done once in eigenpy. Doing it again would allow finding another interpreter, and breaking everything.

src/CMakeLists.txt
```diff
@@ -189,6 +189,11 @@ add_library(${LIBRARY_NAME}
   )
 add_library(${LIBRARY_NAME}::${LIBRARY_NAME} ALIAS ${LIBRARY_NAME})
 
+if(COAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL)
+  add_library(hpp-fcl ALIAS ${LIBRARY_NAME})
+  add_library(hpp-fcl::hpp-fcl ALIAS ${LIBRARY_NAME})
+endif()
```
nim65s, line 195: this is already done in hpp-fclTargets.cmake

nim65s (changes requested):
Some more extensive tests show that a couple fixes are required. I pushed those on https://github.com/lmontaut/hpp-fcl/pull/4.

lmontaut:
> I'd prefer if we could decouple the workspace change from the rename change. This PR is already impossible to review.

Yes it's my bad, sorry. I should have waited for this PR to be merged to merge @jorisv's changes. @nim65s I can revert the changes if you want.

CMakeLists.txt
jorisv: Python is a direct dependency of our project. Taking a direct dependency transitively is not a good behavior in my point of view.
Also, FINDPYTHON set the guard  `FINDPYTHON_ALREADY_CALLED` that will avoid running findpython again in eigenpy.

src/CMakeLists.txt
jorisv: It's mandatory because of the workspace feature. Our project must setup the same targets than the hpp-fcl module. This allow also to use our project via find_package or add_subdirectory without modifying consumer.

src/CMakeLists.txt
jorisv: I don't get in what commit do you find this line.
In the final merge, the alias is only setup if COAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL is set.

python/CMakeLists.txt
jorisv: I'm pretty sure not all commit from this PR will create a stable/buildable project. So do we really care ?

src/CMakeLists.txt
nim65s: Ok, now I understand, thanks. This code path would not be interpreted in the same use cases as hpp-fclTargets.cmake.

nim65s:
We should probably push the topic/coal branch to this repo, to have easier ways of making PR to that branch. I had to make some to your fork instead, which make the history harder to follow. I think this is fine for now, we can avoid the extra cost of undoing and re-doing again. But next time, I really think we need to use a cleaner workflow

PR #598, opened 2024-06-24
Switch to the system dependency for octomap.

clalancette:
This ensures that we have a consistent ABI between this package that uses octomap, and anything else in the system that might use it.  We'll eventually be remove the octomap that we currently have vendored within ROS itself so we can ensure this.

This was enabled by the merging of https://github.com/ros/rosdistro/pull/41623


nim65s (approved):
Thanks !

wxmerkt:
@jcarpent we need a new release now before merging #596 to avoid the ROS released versions becoming orphaned and broken. Could you please create one and tag me? I'll get fixed releases out.

jcarpent:
@wxmerkt I've made a new release. Please see [v.2.4.5](https://github.com/humanoid-path-planner/hpp-fcl/releases/tag/v2.4.5)

PR #601, opened 2024-06-26
Fix a large bunch of warnings

wxmerkt:
Won't get us to zero but will reduce some of the recent additions


nim65s (approved):

src/narrowphase/support_functions.cpp
```diff
@@ -132,12 +132,13 @@ void getShapeSupport(const TriangleP* triangle, const Vec3f& dir,
...
   }
 }
-getShapeSupportTplInstantiation(TriangleP);
+getShapeSupportTplInstantiation(TriangleP)
 
-// ============================================================================
-template <int _SupportOptions>
-inline void getShapeSupport(const Box* box, const Vec3f& dir, Vec3f& support,
-                            int& /*unused*/, ShapeSupportData& /*unused*/) {
+    // ============================================================================
```
wxmerkt, line 137: Hmm, given the changes pre-commit is doing here this change may actually be unfavourable. I didn't see this :-( Would it be worth turning off Wpedantic then?

nim65s: we can surround those with `// clang-format off` and on

wxmerkt: Thanks for the reminder - I did that but it didn't work unless I did something wrong?

nim65s: This seems to work:
```cpp
// clang-format off
template <int _SupportOptions>
inline void getShapeSupport(const Box* box, const Vec3f& dir, Vec3f& support,
                            int& /*unused*/, ShapeSupportData& /*unused*/) {
  // clang-format on
  ```
  Another solution is to cheat by removing the last semicolon inside the macro, so that we can add a semicolon at the end of the macro call, I guess.

jcarpent:
@lmontaut How do you want to proceed?

lmontaut:
@wxmerkt Thanks a lot for fixing all these warnings!
@jcarpent I think we can merge this PR, I will take care of rebasing #596.

PR #611, opened 2024-07-19
Fix serialization unit test when running without Qhull support

jorisv:
Don't run test on hpp::fcl::Convex without Qhull.

This should fix the ROS CI build.


lmontaut (approved):

wxmerkt:
@nim65s Can we have a patch release with these changes before the renaming happens please?

Issue #590, opened 2024-06-07
The o1 and o2 attributes of the hppfcl.Contact class return objects at different addresses in Python


jongyaoY:
I am trying to retrieve the collision object names from the collision results by querying a dictionary created as:

```python
self._names={}
geom = fcl.BVHModelOBB()
geom.beginModel(len(mesh.faces),
                    len(mesh.vertices))
geom.addSubModel(verts, faces)
geom.endModel()
self._names[id(geom)] = name
```
after I get the collision results from the collision manager, I would like to query self._names for the mesh names:

```python
self._manager = fcl.DynamicAABBTreeCollisionManager()
self._manager.setup()

...register objects...

callback = fcl.CollisionCallBackDefault()
self._manager.collide(callback)
contacts = callback.data.result.getContacts()
for contact in contacts:
    names = (self._name[id(contact.o1)],
                  self._name[id(contact.o1)])

```
and I found out `id(contact.o1)` and `id(contact.o2)` change every time I query them:

```python
print(id(contact.o1)) #>>>139658153979680
print(id(contact.o1)) #>>>139658153977104
```
which makes it impossible to retrieve the mesh names

jcarpent:
This behavior is expected, as `contact.o1` and `contact.o2` both refer to C++ pointers.
We will add helpers to extract a unique id, related to the memory address of the C++ pointee.

jongyaoY:
> This behavior is expected, as `contact.o1` and `contact.o2` both refer to C++ pointers. We will add helpers to extract a unique id, related to the memory address of the C++ pointee.

Thanks, then what is the current way of retrieving which objects in the manager that o1 and o2 are referring to?

jcarpent:
Are you able to install hpp-fcl from source?

jongyaoY:
> Are you able to install hpp-fcl from source?
Yes, I complied the package from source. But it didn't resolve the issue. I figured the issue comes from eigenpy, and I noticed in the [CMakeLists.txt](https://github.com/humanoid-path-planner/hpp-fcl/blob/devel/CMakeLists.txt), it's still using the old version of eigenpy. How can I compile it with the latest eigenpy release?

jcarpent:
Please install with conda from a clean env.

jongyaoY:
> Please install with conda from a clean env.

Is there way to compile with cmake and still get the latest eigenpy? since conda takes forever to install

nim65s:
[...]

PR #618, opened 2024-08-21
fix issue #590

jongyaoY:
Solved [#590] by exposing IdVisitor of collision geometries, but it requires eigenpy >=3.7.0


nim65s (changes requested):
This require a new version of eigenpy, v3.8.0, so this is a breaking change.
It will break at least the CI on systems were eigenpy v3.8.0 is not yet available.

Could you guard those changes with `#if EIGENPY_VERSION_AT_LEAST(3, 8, 0)`, to fix this ?

nim65s (approved):
Thanks !

PR #633, opened 2024-11-29
Fix hpp-fclConfig.cmake on Windows

jorisv:
Use IMPLIB instead of LOCATION on Windows


nim65s (dismissed):
Should be good enough, as long as nobody try to patch the install location of this file, between paths that have a different number of subdirs **AND** want to keep the compat layer.

jcarpent (approved):

nim65s (approved):

PR #641, opened 2024-12-30
Support python 3.13

Moisan:
* Replace Python 3.12 by Python 3.13 as the default
* Update pixi.lock
* Test Python 3.13 in CI

### DIFF
--- a/.github/workflows/macos-linux-pip.yml
+++ b/.github/workflows/macos-linux-pip.yml
@@ -40,7 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu", "macos"]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
         exclude:
           - os: "macos"
             python-version: "3.8"  # Not available on arm64
--- a/pixi.toml
+++ b/pixi.toml
@@ -97,9 +97,11 @@ dependencies = { octomap = ">=1.10" }
 
 [feature.py313.dependencies]
 python = "3.13.*"
+scipy = ">=1.14.1"
 
 [feature.py39.dependencies]
 python = "3.9.*"
+scipy = "<1.14.0"
 
 # Use clang-cl on Windows.
 # We must use scripts instead of env to setup CC and CXX
--- a/pixi.toml
+++ b/pixi.toml
@@ -95,8 +95,8 @@ activation = { env = { COAL_HAS_QHULL = "ON" } }
 [feature.octomap]
 dependencies = { octomap = ">=1.10" }
 
-[feature.py312.dependencies]
-python = "3.12.*"
+[feature.py313.dependencies]
+python = "3.13.*"
 
 [feature.py39.dependencies]
 python = "3.9.*"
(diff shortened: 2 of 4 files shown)


.github/workflows/macos-linux-pip.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu", "macos"]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
```
nim65s, line 43: This is not ready yet for pip

nim65s (changes requested):

Moisan:
I see that `scipy`, which is a transitive dependency through `eigenpy`, started supporting Python 3.13 in [release `4.14.1`](https://github.com/scipy/scipy/releases/tag/v1.14.1) and dropped support for python 3.9 in [release `4.14.0`](https://github.com/scipy/scipy/releases/tag/v1.14.0). Hence, if we want to support Python 3.13, we would need to drop Python 3.9 support.

nim65s:
no, we can depend on scipy 4.14.1 for python 3.13 and on scipy 4.13 for python 3.9

pixi.toml
```diff
@@ -95,11 +95,13 @@ activation = { env = { COAL_HAS_QHULL = "ON" } }
 [feature.octomap]
 dependencies = { octomap = ">=1.10" }
 
-[feature.py312.dependencies]
-python = "3.12.*"
+[feature.py313.dependencies]
```
jorisv, line 98: Maybe we can rename this feature into py-latest and the py39 into py-oldest ?

pixi.toml
```diff
@@ -95,11 +95,13 @@ activation = { env = { COAL_HAS_QHULL = "ON" } }
...
+python = "3.13.*"
+scipy = ">=1.14.1"
```
jorisv, line 100: I think you don't need this constraint.
I run without it locally and pixi resolve the good version.

pixi.toml
```diff
@@ -95,11 +95,13 @@ activation = { env = { COAL_HAS_QHULL = "ON" } }
...
 
 [feature.py39.dependencies]
 python = "3.9.*"
+scipy = "<1.14.0"
```
jorisv, line 104: Same here, pixi use the 1.13.1 verison without setting this constraint.

jorisv (changes requested):

jorisv (approved):

.github/workflows/macos-linux-pip.yml
Moisan: I'm not sure what you mean. What is missing here?

jorisv: I think the pip packages are not publish on python 3.13

jorisv:
@Moisan I have rebase your branch on devel.

.github/workflows/macos-linux-pip.yml
jorisv: @nim65s PIP CI is green even in 3.13. Can we close this discussion ?

nim65s: yes

jorisv (approved):

jorisv (dismissed):

jorisv (approved):

.github/workflows/macos-linux-pip.yml
nim65s, line 43:
```suggestion
        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```

python 3.8 does not work on github macos arm runners

nim65s: Oh, wait, it is already excluded

nim65s (approved):
lgtm for pip part

jorisv:
Merged.

@Moisan thank's for the contribution.

PR #651, opened 2025-01-24
Fix install version.

abussy-aldebaran:
This will install libcoal.so.3.0.0 and libcoal.so as a symlink, instead of just libcoal.so as a binary.


jorisv:
Hello @abussy-aldebaran,

Thank's for the contribution.
Can you add a changelog entry ?

jorisv (approved):

PR #659, opened 2025-02-17
Introduce next-generation Python bindings using `nanobind`

ManifoldFR:
This (draft) pull request adds a second set of Python bindings using the modern [nanobind](https://github.com/wjakob/nanobind) bindings generator.

The new nanobind-based API can be imported as a replacement for the old one by importing:
```python
import coal
```

Only **one** of the bindings can be built at one time. The nanobind bindings can be built by setting the new `COAL_PYTHON_NANOBIND` option to ON.
```bash
cd <build-dir>
cmake . -DCOAL_PYTHON_NANOBIND=ON
```

~~These new bindings depend, at runtime, on the new [nanoeigenpy](https://github.com/Simple-Robotics/nanoeigenpy) support library, which we use to expose Eigen's Quaternion type.~~

# Task checklist 

- [x] Ensure no collision with previous set of Boost.Python bindings:
  - [x] cmake: Move Boost.Python bindings' code from main `CMakeLists.txt` to legacy `python/CMakeLists.txt`
  - [x] cmake: add flag `COAL_PYTHON_BINDINGS`
- [x] Stubs
- [x] Re-expose the following features from the Boost.Python bindings:
  - [x] Version
  - [x] Broadphase (`python/broadphase/broadphase.cc`)
  - [x] GJK (`python/gjk.cc`)
  - [x] MeshLoader
  - [x] Maths (`python/math.cc`)
  - [x] Pickling
  - [x] Serialization
  - [x] Contact patches
  - [x] Distance API (`python/distance.cc`)
  - [x] Octree (`python/octree.cc`)
  - [x] Collision (`python/collision.cc`)
  - [x] Collision Geometries (`python/collision-geometries.cc`)
    - [x] Base class
    - [x] BVH models
    - [x] Heightfields
    - [x] `CollisionObject`
- [x] Test everything


florent-lamiraux:
For your information, after a quick test based on https://stackoverflow.com/questions/8233252/boostpython-and-weak-ptr-stuff-disappearing, nanobind seems as bad as boost python to handle weak pointers.

ManifoldFR:
> For your information, after a quick test based on https://stackoverflow.com/questions/8233252/boostpython-and-weak-ptr-stuff-disappearing, nanobind seems as bad as boost python to handle weak pointers.

Thanks, this is good to know in the future. Almost certainly, weak pointers will be an object management nightmare whatever the bindings generator is.

Anyhow, there is a definite advantage of upgrading our bindings, namely simplifying maintenance, getting modern features (and up-to-date docs), eliminating the Boost.Python dependency, ability to use the Python 3.12+ stable ABI - which are things Boost.Python simply doesn't have and never will 😃

ManifoldFR:
Change of plans: following discussions with @jorisv and @nim65s, the new bindings shall *not* coexist with the legacy Boost.Python bindings. Users will be able to build/install either one, but not both.

**Edit** I will require some help to add integrate the nanobind option to the CI pipelines.

ManifoldFR:
Commits from @Lucas-Haubert got merged in :)

python-nb/pickle.hh
```diff
@@ -0,0 +1,34 @@
...
+    using namespace nb::literals;
+    cl.def("__getstate__", [](const T &obj) -> std::string {
+        std::stringstream ss;
+        boost::archive::text_oarchive oa(ss);
+        oa & obj;
+        return ss.str();
+      }).def("__setstate__", [](T &obj, const std::string &state) {
+      std::istringstream is(state);
+      boost::archive::text_iarchive ia(is, boost::archive::no_codecvt);
+      new (&obj) T();
```
ManifoldFR, line 30: @jcarpent this line was required to avoid a memory read bug. It seems like a difference with Boost.Python is that nanobind does _not_ allocate the object passed to `__setstate__`, so we have to placement-new it.

I don't feel comfortable with placement-new'ing using the default constructor because it might not always exist. It works for now because all the objects we allow pickling for have default constructors. Do you know if Boost.Serialization could handle object construction for us?

Lucas-Haubert:
Hello @nim65s . The first batch of commits is done here. After discussing with @ManifoldFR, we wondered if you could adapt the pip workflow regarding the intergration of nanobind.

nim65s:
sure

ManifoldFR:
I've just rebased on devel.

ManifoldFR:
Actually rebasing on `devel` failed because of recent (major?) changes to the core Coal library. @Lucas-Haubert can you take a look? I think #682 is the culprit.

To make this easier you can just tack on some additional post-rebase fix commits.

ManifoldFR:
Oh my God I hate Windows so much

python-nb/fwd.h
ManifoldFR, line 1: @Lucas-Haubert we forgot to put include guards here.. and other headers for the nanobind bindings as mentioned by the great @jorisv

ManifoldFR:
Only thing left is to fix the ROS build.

@nim65s would you have time for that, this week?

nim65s:
Yes, you can have a look at https://github.com/ManifoldFR/coal/pull/2.
This seems ok for Nix, but I'm afraid for ROS we'll have to wait

ManifoldFR:
I think that, while we are waiting https://github.com/ros/rosdistro/pull/45744 to be merged, we can still merge this PR.

Further steps, in other PRs:
- port doxygen autodoc ?
- write-up a porting guide detailing our experience, perhaps some benchmarks, pursuant to https://github.com/wjakob/nanobind/discussions/1001

ManifoldFR:
I've just rebased on `devel`. (Highly likely I might have broken nix or something here)

ping @nim65s

ManifoldFR:
Following #724, we can expose the added overload in nanobind using `<nanobind/stl/function.h>`.

ManifoldFR:
I made an empty merge commit of `devel` into this branch due to a lockfile conflict following #729

I've exposed the function-based overloads for `broadphase.collide()` and `broadphase.distance()`.

ManifoldFR:
With this our work is done. Only two things are the docstrings for the bindings (will be done later), and the ROS CI which is snafu

Ping @nim65s @jorisv

CMakeLists.txt
```diff
@@ -162,12 +162,22 @@ APPLY_DEFAULT_APPLE_CONFIGURATION()
 
 option(BUILD_PYTHON_INTERFACE "Build the python bindings" ON)
 cmake_dependent_option(
-  GENERATE_PYTHON_STUBS
-  "Generate the Python stubs associated to the Python library"
-  OFF
+  COAL_PYTHON_NANOBIND
+  "Build the nanobind-based bindings."
+  ON
```
nim65s:
```suggestion
  OFF
```

Activating this by default would be an unnecessary breaking change. We can do that on the next major release.

CMakeLists.txt
```diff
@@ -512,7 +513,28 @@ endif(COAL_HAS_OCTOMAP)
...
+    ${PROJECT_NAME}_python
+    PROPERTIES EXCLUDE_FROM_DEFAULT_BUILD True
+  )
+  if(COAL_PYTHON_NANOBIND)
+    if(CMAKE_VERSION VERSION_LESS 3.18)
+      set(DEV_MODULE Development)
+    else()
+      set(DEV_MODULE Development.Module)
+    endif()
+    find_package(Python REQUIRED COMPONENTS Interpreter ${DEV_MODULE})
```
nim65s:
```suggestion
    find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
```



we can target CMake 3.22, ref. https://github.com/jrl-umi3218/jrl-cmakemodules#supported-cmake-versions

nim65s (changes requested):

CMakeLists.txt
ManifoldFR: Oh yes, we've removed support for the old ones since this PR was initially created! Will fix this asap. Thanks

nim65s (approved):

ManifoldFR:
Thanks @nim65s 😃 

And thanks to everyone who worked on this @Lucas-Haubert @jorisv @nim65s, the first step in modernising the bindings in our entire stack 🎉

jorisv (approved):

ManifoldFR:
❤️ ❤️ ❤️

PR #668, opened 2025-03-12
Add tracy profiling

lmontaut:
- added cmake option `COAL_BUILD_WITH_TRACY`
- put tracy scoped zones in broadphase and primitive shapes collision/distance queries

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,14 +179,6 @@ if(COAL_BUILD_WITH_TRACY)
     Tracy::TracyClient
     PROPERTIES POSITION_INDEPENDENT_CODE True
   )
-  if(${Tracy_FOUND})
-    message(STATUS "Tracy found on your system at ${Tracy_DIR}")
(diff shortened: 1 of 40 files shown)


CMakeLists.txt
```diff
@@ -165,6 +171,24 @@ cmake_dependent_option(
...
+    PROPERTIES POSITION_INDEPENDENT_CODE True
+  )
+  if(${Tracy_FOUND})
+    message(STATUS "Tracy found on your system at ${Tracy_DIR}")
+  else()
+    message(
+      FATAL_ERROR
+      "Coal support for tracy is enabled, but tracy was not found on your system."
+    )
+  endif()
```
jcarpent: I don't think it is necessary to add this line as we have already `ADD_PROJECT_DEPENDENCY(Tracy REQUIRED)`. Could you remove these lines?

jcarpent (dismissed):

PR #697, opened 2025-04-25
Fix autodoc generation on Windows

jorisv:
Windows build with python autodoc was not working because two function with the same signature were generated.

By using the same return type this PR fix this issue.


lmontaut (approved):

jorisv:
Since the regression had not been released no need to update the changelog.

PR #702, opened 2025-05-02
[cmake] Remove superfluous call to `FINDPYTHON`

ManifoldFR:
+ eigenpy already does it for us


jorisv (changes requested):
I will check it soon

ManifoldFR:
Following discussion with @jorisv I will be closing this. Another fix will be pushed on #659.

jorisv:
I can't reproduce this issue so we will keep the FINDPYTHON call for now.
We will wait jrl-cmakemodule v2 to remove it.

PR #709, opened 2025-05-26
CMake: add COAL_DISABLE_HPP_FCL_WARNINGS option

nim65s:


jmirabel:
Thank you !

hpp-fclConfig.cmake
```diff
@@ -24,4 +22,13 @@ if(NOT TARGET hpp-fcl::hpp-fcl)
       INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
       INTERFACE_LINK_LIBRARIES "coal::coal"
   )
+
+  if(COAL_DISABLE_HPP_FCL_WARNINGS)
+    target_compile_definitions(
+      hpp-fcl::hpp-fcl
+      INTERFACE -DCOAL_DISABLE_HPP_FCL_WARNINGS
```
jorisv:
```suggestion
      INTERFACE COAL_DISABLE_HPP_FCL_WARNINGS
```

No need to add `-D`: https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

src/CMakeLists.txt
```diff
@@ -222,6 +222,12 @@ if(COAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL)
     ${LIBRARY_NAME}
     PUBLIC -DCOAL_BACKWARD_COMPATIBILITY_WITH_HPP_FCL
   )
+  if(COAL_DISABLE_HPP_FCL_WARNINGS)
+    target_compile_definitions(
+      ${LIBRARY_NAME}
+      PUBLIC -DCOAL_DISABLE_HPP_FCL_WARNINGS
```
jorisv:
```suggestion
      PUBLIC COAL_DISABLE_HPP_FCL_WARNINGS
```

jorisv (changes requested):

nim65s:
Yeah, I agree, but I just did what the others around where doing.
Mind if I fix the others too in this PR then ?

jorisv:
You can remove other -D.

I raised another issue in the pinocchio PR. Did you think it's a good idea to forward this definition in the target ? If we do so the packager will choose if we display the warning or not. I think it's better to leave end use choose by defining the define manually in there own project.

We can display the define name in the warning message.

jorisv (dismissed):

jorisv (approved):

PR #716, opened 2025-06-12
convex: add missing typedefs for backward compatibility

lmontaut:


ManifoldFR:
thanks for this @lmontaut :)

jorisv (approved):

Issue #725, opened 2025-06-23
ContactPatchResult Native exception: vector too long


hazidh:
I wrote the following code and run it, it gives `std::length_error`, I suspect it is the ContactPatchResult `set(const ContactPatchRequest& request)`
```
CollisionResult Cone_Sphere_CollisionConatactPatchTest(double sphereRadius, 
    double coneHeight, double coneRadius,
    Transform3s T1, Transform3s T2, 
    CollisionRequest col_req, CollisionResult col_res) {

    //std::shared_ptr<coal::Sphere> shape1 = std::make_shared<coal::Sphere>(sphereRadius);
    std::shared_ptr<coal::Box> shape1 = std::make_shared<coal::Box>(2,2,2);
    std::shared_ptr<coal::Box> shape2 = std::make_shared<coal::Box>(1,1,1);
    /*coal::BVHModel<coal::AABB>* shape1 = new coal::BVHModel<coal::AABB>;
    cteateBVH(shape1, 1,1,1);
    std::shared_ptr<coal::Cone> shape2 = std::make_shared<coal::Cone>(coneHeight, coneRadius);*/

    //collide(shape1, T1, shape2.get(), T2, col_req, col_res);
    collide(shape1.get(), T1, shape2.get(), T2, col_req, col_res);
    std::cout << "ContactReq_NUM:" << col_req.num_max_contacts << std::endl;
    std::cout << "ContactRes_NUM:" << col_res.numContacts() << std::endl;

    ContactPatchRequest patch_req(col_req);
    ContactPatchResult patch_res(patch_req);

    computeContactPatch(shape1.get(), T1, shape2.get(), T2, col_res, patch_req, patch_res);
    
    std::cout << "Contact Patch Num:" << patch_res.numContactPatches() << std::endl;
    for (int i = 0; i < patch_res.numContactPatches(); i++) {
        for (int j = 0; j < patch_res.getContactPatch(i).points().size(); j++) {
            std::cout << "Contact Point :" << patch_res.getContactPatch(i).getPoint(j).transpose() << std::endl;
        }
    }
    return col_res;

}
int main() {


    /*
    * Test the collision
    */
    CollisionResult col_res;
    CollisionRequest col_req(CollisionRequestFlag::CONTACT, 9223372036854775807);

[...]

PR #742, opened 2025-08-08
Fix ContactPatchResult falling back to native exception for excessive…

andrewleachtx:
… contact request size

  Addresses #725 which produces `std::length_error` instead of a clear error for when a user exceeds a reasonable
  size for contact points in a collision request.

  Modifies `ContactPatchResult::set(const ContactPatchRequest& request)` to check according to a const.

  Added `contact_path_excessive_requests` test case to verify proper exception handling for both reasonable and
  excessive contact patch requests.

### DIFF
--- a/include/coal/collision_data.h
+++ b/include/coal/collision_data.h
@@ -916,6 +916,19 @@ struct COAL_DLLAPI ContactPatchResult {
 
   /// @brief Set up a `ContactPatchResult` from a `ContactPatchRequest`
   void set(const ContactPatchRequest& request) {
+    // In practice there may be better constants, and coal may have one. But for
+    // contact points, this should suffice for memory limits on most systems.
+    const size_t max_reasonable_patches = 100000;
+
+    // Check if the request exceeds reasonable limits before attempting resize
+    if (request.max_num_patch > max_reasonable_patches) {
+      COAL_THROW_PRETTY(
+          "Requested too many contact patches: " +
+              std::to_string(request.max_num_patch) +
+              ". Maximum supported: " + std::to_string(max_reasonable_patches),
+          std::invalid_argument);
+    }
+
     if (this->m_contact_patches_data.size() < request.max_num_patch) {
       this->m_contact_patches_data.resize(request.max_num_patch);
     }
--- a/include/coal/collision_data.h
+++ b/include/coal/collision_data.h
@@ -918,7 +918,8 @@ struct COAL_DLLAPI ContactPatchResult {
   void set(const ContactPatchRequest& request) {
     // In practice there may be better constants, and coal may have one. But for
     // contact points, this should suffice for memory limits on most systems.
-    const size_t max_reasonable_patches = 100000;
+    const size_t max_reasonable_patches =
+        this->m_contact_patches_data.max_size();
 
     // Check if the request exceeds reasonable limits before attempting resize
     if (request.max_num_patch > max_reasonable_patches) {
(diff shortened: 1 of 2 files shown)


include/coal/collision_data.h
```diff
@@ -916,6 +916,19 @@ struct COAL_DLLAPI ContactPatchResult {
 
   /// @brief Set up a `ContactPatchResult` from a `ContactPatchRequest`
   void set(const ContactPatchRequest& request) {
+    // In practice there may be better constants, and coal may have one. But for
+    // contact points, this should suffice for memory limits on most systems.
+    const size_t max_reasonable_patches = 100000;
+
+    // Check if the request exceeds reasonable limits before attempting resize
+    if (request.max_num_patch > max_reasonable_patches) {
```
jorisv, line 925: You can maybe use [std::vector::max_size](https://en.cppreference.com/w/cpp/container/vector/max_size.html) instead of a magic constant.

jorisv (changes requested):
Thank you for the PR.

@lmontaut Do you agree with this change ?

include/coal/collision_data.h
andrewleachtx: I added a newer commit, I like that. Before I didn't really know what was reasonable, so I used `100,000`.

My concern is that each `ContactPatch` stored in `m_contact_patches_data` has its own heap allocated data (a `std::vector<Vec2s>`), so I don't know what a better bound is. Any suggestions? Is it tractable?

I guess that is chasing possible errors outside of the scope of this issue, which is simply caused by an excessive resize.

jorisv: Here, we protect against a miss use of `std::vector::resize`. But this issue can happen on each std::vector (by calling `resize`, `reserve` or too many time `push_back`).

If this issue happen frequently on some vectors we can implement one of the following option:

- Throw an exception with a nice message
  - User know what happen
  - The program is stopped and user don't have many option to solve the issue   
- Log an error message and ignore new patch if there is more than `MAX_USER_CONFIG_PATCHES` 
  - User know what happen
  - User can setup a memory upper bound
  - The program is not stopped but user know something sub-optimal is happening 
- Instead of logging, allow user to setup a callback

We can also watch many vector. (if all allocated data by `ContactPatchResult::set` are higher than `MAX_USER_CONFIG_TOTAL_PATCHES_BYTES`).

I think we must wait library owner feedback @lmontaut. We should have some by the end of the month.

lmontaut:
Hey @andrewleachtx, apologies for the delayed answer, I was on vacation.
Thanks a lot for your contribution!
As you said in #725, the problem stems from the user creating a `CollisionRequest` with an ungodly amount of contacts. Indeed, `100000` maximum contacts should be **way more** than enough for most applications out there. 
The `set` method of contact patches is not the culprit here, but rather (the user and) the constructor of `CollisionRequest`. So, if a check were to be done, it should be done in this constructor. Once the user has set the maximum number of contacts, coal never touches this information and never resizes contact data; it thus creates as many potential contact patches as this maximum number of contact points decided by the user.

That being said, @jorisv and I discussed, and we believe it's up to the end user to make the right decision when it comes to deciding on the maximum number of contacts. We could add this check, but then we would have to check a lot more user inputs. Little by little, this adds more stuff to maintain.

andrewleachtx:
Sounds good and no worries @lmontaut. Thanks for your feedback. Definitely agree that it feels like a problem that would just keep growing and growing if you try to consistently enforce this same rule everywhere!

PR #753, opened 2025-09-08
Fix sqrDistLowerBound in octree traversal with height field

xndcn:


include/coal/internal/traversal_node_octree.h
```diff
@@ -641,8 +641,11 @@ class COAL_DLLAPI OcTreeSolver {
                         (int)Contact::NONE, c1, c2, -normal, distance));
           }
         }
-      } else
-        sqrDistLowerBound = distToCollision * distToCollision;
```
lmontaut, line 645: From my understanding, `sqrDistLowerBound` is only used in [`collide`](https://github.com/coal-library/coal/blob/5f29ba66996da52707722ac4ba42168357fc86f8/src/collision_node.cpp#L61) (involving `CollisionTraversalNodeBase`) to do sanity checks (see [checkResultLowerBound](https://github.com/coal-library/coal/blob/5f29ba66996da52707722ac4ba42168357fc86f8/src/collision_node.cpp#L73)). The actual lower bound of the squared distance is stored in `CollisionResult` (e.g. as done [here](https://github.com/coal-library/coal/blob/5f29ba66996da52707722ac4ba42168357fc86f8/include/coal/internal/traversal_node_octree.h#L649)).

@xndcn why do you want to make these modifications?

lmontaut (changes requested):

include/coal/internal/traversal_node_octree.h
xndcn: @lmontaut Thanks. Yes, the change is a defensive fix for sanity checks. `sqrDistLowerBound` is a shared variable reused across recursive calls, one call can set it to 0 when a collision is found, and a later recursive branch can overwrite it with a non-zero value. Without an extra comparison, we can end up with a non-zero `sqrDistLowerBound` with collision and trigger the assert in `checkResultLowerBound`.

lmontaut: Yes, you are right. To be consistent with the whole library, I think your modification needs to be propagated not only to `octree-heighfield` but to all collision pairs using a `CollisionTraversalNodeBase`. @xndcn do you want to handle that or do you want me to do it?

(This means modifying all `leafCollides` functions.)

xndcn: @lmontaut After reviewing `leafCollides` and `collisionRecurse`, my understanding is that if `leafCollides` does not contain inner recursive calls, then `sqrDistLowerBound` can safely be managed within `collisionRecurse`. The issue mainly arises with `leafCollides` in `OcTree-*`, since all of them perform recursive calls internally. Among these, all `OcTree-*` implementations except `OcTree-HeightField` already handle `sqrDistLowerBound` correctly by codes like
```
sqrDistLowerBound = std::max((Scalar)0, result->distance_lower_bound);
sqrDistLowerBound *= sqrDistLowerBound;
```

lmontaut: You're right about Octrees, but it seems like this is not the case with BVHs for example. BVHs can contain recursive calls to `collisionRecurse`. Therefore [this line in BVH traversal](https://github.com/coal-library/coal/blob/5f29ba66996da52707722ac4ba42168357fc86f8/include/coal/internal/traversal_node_bvhs.h#L231) can cause issues. Do you agree @xndcn ?

xndcn: @lmontaut I'm not sure, but I think if BVHs do not contain `self-recursive` calls like `octree-heighfield`, then `sqrDistLowerBound` can be handled in `collisionRecurse`. In such cases, the code already ensures correctness with logic like:
```
sqrDistLowerBound = std::min(sqrDistLowerBound1, sqrDistLowerBound2);
```

lmontaut: Ah yes you're right @xndcn. Thanks for your contribution! :)

lmontaut (approved):

PR #758, opened 2025-09-25
Optimize the linear and log convex shape support function

jorisv:
This PR introduce the [highway framework](https://github.com/google/highway/tree/master)

# TODO

- [x] Add linear support function benchmark
- [x] Optimize highway implementation
- [x] Add log support function benchmark
- [x] Find log support function efficient implementation
- [x] Use SparseSet to improve visited node list clear
- [ ] Look at graph reordering algorithm
- [ ] Replace current implementation by optimized one

# Discussion

- [x] Should highway be an optional dependency: Yes
- [x] Should we hide highway from the public API: Yes
- [ ] Should we constraint the number of architectures ?


benchmark/convex_support.cpp
```diff
@@ -0,0 +1,356 @@
...
+#include <vector>
+#include <cmath>
+#include <iostream>
+
+static Eigen::Vector3d toSphere(const Eigen::Vector3d& point, double scale) {
+  return scale / point.norm() * point;
+}
+
+struct Icosahedron {
+  using TriangleIndex = std::tuple<std::size_t, std::size_t, std::size_t>;
```
ManifoldFR: Je pinaille, but `std::array<std::size_t, 3>` would be more appropriate here no?

PR #770, opened 2025-10-15
CMake: lib version/soversion

nim65s:
Hi,

My main goal with this is to improve compatibility between different pip minor releases. Thanks @Gregwar for the initial report.

But in general: I think we need to discuss a global policy for this on all our packages.

ref. https://github.com/stack-of-tasks/eigenpy/blob/73265595950774908d1555636b5ab79383e8c8ed/CMakeLists.txt#L59 for the option name and its initial value


jorisv:
Hello,

If I understand well this will create a symlink named libcoal.so.3 instead of libcoal.so ?

nim65s:
Currently we have a `libcoal.so.3.0.2` lib, and a `libcoal.so` symlink to it.

So if eg. pinocchio is linked to `libcoal.so.3.0.1` and a package manager update to `libcoal.so.3.0.2` and remove `libcoal.so.3.0.1`, pinocchio is broken for wrong reasons.

A more common convention is to specify both `VERSION` and `SOVERSION` such that `SOVERSION` matches the first component of `VERSION`, so we would have `libcoal.so.3` and a `libcoal.so` symlink to it.

ref. https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html

I think we should allow this behavior to be desactivable through a CMake option, and I copy/pasted the existing eigenpy config for this.

jorisv:
Ok, I was thinking we were linking against the full version because we never guarantee ABI compatibility.

jcarpent (dismissed):

jorisv (approved):

PR #813, opened 2026-02-24
Enhance patch simplifier

lmontaut:
- Improves the `ContactPatchSimplifierMaxArea` algorithm.
- Adds better tests for patch simplifiers


include/coal/utils.h
```diff
@@ -0,0 +1,148 @@
...
+#include <algorithm>
+
+namespace coal {
+
+// -----------------------------------------------------------
+// Collection of various utilities useful in the Coal library.
+// -----------------------------------------------------------
+
+/// @brief Compute the convex hull of a polygon.
+inline void computePolygonConvexHull(const std::vector<Vec2s>& cloud,
```
jorisv: This code is not templated and is quiet complex. I think it's better to put it in a .cpp file.

include/coal/utils.h
```diff
@@ -0,0 +1,148 @@
...
+  }
+  Vec2s tmp = cloud_[0];
+  cloud_[0] = cloud_[support_idx];
+  cloud_[support_idx] = tmp;
+  cvx_hull.clear();
+  cvx_hull.emplace_back(cloud_[0]);
+  const Vec2s& v = cvx_hull[0];
+
+  // Step 2 - Sort the rest of the point cloud_ according to the angle made with
+  // v. Note: we use stable_sort instead of sort because sort can fail if two
```
jorisv, line 75: I don't think it's goodly worked. `std::sort` can't fail, but it can switch identical element. Is it important for this algorithm ?

include/coal/utils.h
```diff
@@ -0,0 +1,148 @@
...
+  if (cloud.size() <= 2) {
+    // Point or segment, nothing to do.
+    for (const Vec2s& point : cloud) {
+      cvx_hull.emplace_back(point);
+    }
+    return;
+  }
+
+  // temporary copy of cloud to work on
+  COAL_MAKE_ALLOCA_ARRAY_VIEW(Vec2s, cloud_, cloud.size());
```
jorisv: Is it safe to call alloca for a convex hull that can be quit big ?
Maybe you should add in the documentation you except `cloud` to be small.

include/coal/container/array_view.h
jorisv, line 1: Should we use boost::span ?
Main issue is boost::span isn't available in Ubuntu 22.04.
We add them manually in [aligator](https://github.com/Simple-Robotics/aligator/pull/383/changes#diff-9d9f3459d4c06c1b1e512085a1abff1af8690cb4e538cbd9b43e75e07642b19f) when building in ubuntu 22.04.

include/coal/utils.h
jorisv, line 1: Do you think it's a good naming ? Maybe `include/coal/contact_patch/convex_hull.h` is better

include/coal/contact_patch/contact_patch_solver.h
```diff
@@ -171,7 +171,7 @@ struct COAL_DLLAPI ContactPatchSolver {
 
   /// @brief Retrieve result, adds a post-processing step if result has bigger
   /// size than `this->max_patch_size`.
-  void getResult(const Contact& contact, const ContactPatch::Polygon* result,
+  void getResult(const Contact& contact, const std::vector<Vec2s>* result,
```
jorisv, line 174: Is this an API break or ContactPatch::Polygon was already a std::vector<Vec2s> ?

jorisv (changes requested):
Update CHANGELOG.md

include/coal/contact_patch/contact_patch_solver.h
lmontaut: ContactPatch::Polygon was already a std::vector

include/coal/utils.h
lmontaut: Agreed. I will transfer this to a new file.

include/coal/utils.h
lmontaut: I will add it in the documentation. I think if people want to use this function with large polygons in the future, we can add a switch that uses either a malloc or an alloca. For now this method is primarly used for relatively small cloud points, so alloca should be fine.

include/coal/container/array_view.h
lmontaut: I can look into it but for now ArrayView is only used in one place. If something like that is needed more broadly in coal in the future, maybe we can switch to boost::span? What do you think?

include/coal/utils.h
lmontaut: I agree. I moved `computePolygonConvexHull` declaration to a new .h file `include/coal/contact_patch/polygon_convex_hull.h`. The implementation is in `src/contact_patch/polygon_convex_hull.cpp`.

include/coal/utils.h
lmontaut: @jorisv can you elaborate?

include/coal/utils.h
lmontaut: I added a `COAL_ALLOCA_MAX_STACK_BYTES` above which alloca switches to a heap allocated pointer.

include/coal/container/array_view.h
lmontaut: Now using boost::span

CMakeLists.txt
```diff
@@ -188,6 +188,19 @@ if(COAL_ENABLE_LOGGING)
   ADD_PROJECT_DEPENDENCY(Boost REQUIRED log)
 endif()
 
+# boost::span requires Boost >= 1.73.0 (released August 2020).
+# On Ubuntu 20.04 the system Boost is 1.71 and is too old; Ubuntu 22.04
+# ships 1.74 and is fine.
```
jorisv: `boost/core/span.hpp` had been introduced in boost 1.78. It's not available on ubuntu 22.04.

jorisv (changes requested):

jorisv (approved):

PR #816, opened 2026-03-06
Fix condition for negative bounding volume check

rjoomen:
A triangle has a volume of zero, but gets flagged here as having a negative volume.


lmontaut (approved):

lmontaut:
Hi @rjoomen thanks for the PR! Side note: the AABB of a triangle doesn't necessarily have a zero volume, only when the triangle belongs to the x, y or z plane.
I agree that the check should be `< 0` rather than `<= 0` to cover the triangle edge cases.

Thanks!

rjoomen:
> Hi @rjoomen thanks for the PR! Side note: the AABB of a triangle doesn't necessarily have a zero volume, only when the triangle belongs to the x, y or z plane.
> 
> I agree that the check should be `< 0` rather than `<= 0` to cover the triangle edge cases.
> 
> 
> 
> Thanks!

Good one, that would only be true for the OBB indeed. Thanks for the heads up.

PR #819, opened 2026-03-17
Add missing calls to computeLocalAABB for internal objects

rjoomen:
When an internal collision object is constructed it needs a local AABB for the GJK BoundingVolumeGuess. In most places computeLocalAABB is correctly called, this PR adds a few missed cases.


lmontaut (approved):

lmontaut:
Thanks @rjoomen !

Issue #792, opened 2025-12-05
New primitive implementation


JonasBreuling:
I'm interested in adding a new primitive to the library. Specifically, a deformed cylinder (with arbitrary oriented caps and possibly different radii; i think this is the convex hull of two oriented disks) as depicted in the figure below.

![Image](https://github.com/user-attachments/assets/0b9d4559-8d66-4387-8cc0-30382dbf8085)

According to [support_functions.cpp](https://github.com/coal-library/coal/blob/devel/src/narrowphase/support_functions.cpp) and [geometric_shapes.h](https://github.com/coal-library/coal/blob/devel/include/coal/shape/geometric_shapes.h) this requires to implement
- the support function `getShapeSupport`
- the support point `getShapeSupportSet`
- the class definition `DeformedCylinder : public ShapeBase`

Open questions:
- what else has to be implemented?
- what is `cylinder->getSweptSphereRadius()`; this is possibly important for a new primitive

If this can be done, I would like to proceed with the convex hull of two oriented rectangles.

lmontaut:
Hey @JonasBreuling, sorry for the late reply!
I asked Claude (see below) to make a summary of what's needed to fully support a shape in Coal and I agree with it.
I'm adding it to this issue for reference, but you can start by doing only a few things:
- class definition `DeformedCylinder : public ShapeBase`
- add a new entry in the `NODE_TYPE` enum (so that internal algos can efficiently dispatch to your shape)
- implement `getShapeSupport`

If you want to use your shape using coal's `collide` API:
- add your shape to the collision functions matrix thanks to the new `NODE_TYPE` entry (you don't have to fill the entire matrix at first, only the pairs you intend to support)
- modify `src/narrowphase/minkowski_difference.cpp` so that GJK's support calls are properly dispatched on your new shape

If you want support for contact patches (this is often more complicated than implementing the support function):
- implement `getShapeSupportSet`
- add your shape to the contact patch functions matrix
- modify `src/contact_patch/contact_patch_solver.cpp` to properly dispatch your new shape's support set function 

Let me know if you have more questions.

```
  Summary: Adding a New Primitive Shape

  1. Define the NODE_TYPE enum entry

  File: include/coal/collision_object.h:64-90

  Add your shape to the NODE_TYPE enum:
  GEOM_YOUR_SHAPE,  // Add before NODE_COUNT

  2. Create the shape class

  File: include/coal/shape/geometric_shapes.h

[...]

PR #822, opened 2026-03-23
Add GEOM_CUSTOM node type for user-defined shapes

rjoomen:
In order to support continuous collision checking in Trajopt (currently done using Bullet), Coal needs support for custom shapes. For Bullet, this was trivial, as btConvexShape can be overridden. Now, using the new GEOM_CUSTOM, this is easy using Coal as well. 
- Details about the new Coal collision manager for Tesseract are here: https://github.com/tesseract-robotics/tesseract_collision_coal/issues/5. Benchmarks show a significant improvements using Coal!
- This also enables an easy way of implementing custom shapes, as discussed here: https://github.com/coal-library/coal/issues/792. I added a test case demonstrating a DeformedCylinder.

As you might notice, a lot of this was created with the help of Claude, for completeness here is the Claude PR message:

Summary

  - Add GEOM_CUSTOM node type and virtual computeShapeSupport() on ShapeBase, enabling user-defined shapes to participate in Coal's GJK/EPA collision, distance, and contact patch pipelines without modifying Coal internals
  - Register GEOM_CUSTOM in all three dispatch matrices (collision, distance, contact patch) for every shape-pair combination, including BVH, HeightField, and OcTree
  - Expose getNormalizeSupportDirection() as public API so wrapper shapes can delegate the Nesterov heuristic to their underlying shape via shape_traits

  Motivation

  Adding a new primitive shape to Coal currently requires modifying the NODE_TYPE enum, registering 63+ dispatch matrix entries, editing the Minkowski difference support dispatch, and adding serialization — all inside Coal's source tree. This was discussed in https://github.com/coal-library/coal/issues/792.

  GEOM_CUSTOM reduces this to three steps for external users:
  1. Subclass ShapeBase
  2. Override getNodeType() to return GEOM_CUSTOM
  3. Override computeShapeSupport() with the shape's support function

  Everything else (GJK, EPA, contact patches, broadphase) works automatically via virtual dispatch.

  Design

  Entirely opt-in. ShapeBase does not default to GEOM_CUSTOM — it inherits BV_UNKNOWN from CollisionGeometry. Custom shapes must explicitly override getNodeType(). Built-in shapes are unaffected; all existing NODE_TYPE values and dispatch paths are unchanged.

  Virtual dispatch only for GEOM_CUSTOM. The dispatch matrices route GEOM_CUSTOM pairs through ShapeShapeCollide<ShapeBase, T> (and equivalents for distance/contact patch), which call computeShapeSupport() via the virtual method. Built-in shapes continue to use their template-specialized fast paths.

  No specialized fast-paths. GEOM_CUSTOM always goes through GJK/EPA. For shapes where closed-form solutions exist (e.g. sphere-sphere), implementing them as a built-in NODE_TYPE will still be faster.

  getNormalizeSupportDirection exposed publicly so that wrapper shapes (e.g. swept-volume shapes for continuous collision detection) can query the Nesterov heuristic for their underlying shape through shape_traits, the single source of truth.

  API additions

  // collision_object.h
  enum NODE_TYPE { ..., GEOM_CUSTOM, NODE_COUNT };

  // geometric_shapes.h — on ShapeBase
  virtual void computeShapeSupport(const Vec3s& dir, Vec3s& support,
                                   int& hint, details::ShapeSupportData& data) const;
  virtual bool needNesterovNormalizeHeuristic() const;

  // minkowski_difference.h
  COAL_DLLAPI bool getNormalizeSupportDirection(const ShapeBase* shape);

  Tests

  Two new test files with 21 test cases:

  custom_shape.cpp (17 tests) — CustomSphere (leaf custom shape) and CastSphere (swept-volume wrapper, demonstrating continuous collision detection):
  - Collision, distance, contact patches vs built-in shapes (Box, Sphere, Capsule, Cylinder, Ellipsoid)
  - Custom-vs-custom pairs
  - BVH mesh collision and distance
  - HeightField collision (distance documents upstream limitation)
  - OcTree collision and distance (both directions)
  - computeBV<AABB/OBB/OBBRSS, ShapeBase> validation
  - Built-in shapes retain their specific NODE_TYPE

  custom_shape_deformed_cylinder.cpp (4 tests) — DeformedCylinder (convex hull of two oriented disks, from https://github.com/coal-library/coal/issues/792):
  - Collision and distance vs Box
  - Degenerates to built-in Cylinder when disks are parallel
  - Contact patch computation

  Test plan

  - All existing Coal tests pass (no regressions)
  - New custom_shape and custom_shape_deformed_cylinder tests pass
  - Built-in shape NODE_TYPE values unchanged (verified by test)


JonasBreuling:
Awesome, thank you @rjoomen !

lmontaut:
Hey @rjoomen thanks a lot for this PR. There will probably be a bit of efficiency loss for custom shapes (because of the virtualization, which can become expensive when calling the support function for example) but I think this feature is very cool for experimentation. I asked @jorisv to have a look into it. Otherwise this looks good to me.

rjoomen:
@lmontaut, @jorisv Friendly ping.

We (as in: the Tesseract Robotics project [repo](https://github.com/tesseract-robotics/tesseract), [docs](https://tesseract-robotics.github.io/tesseract/)) would really like to add Coal as a collision checker. Compared to the currently used Bullet and FCL it has increased performance, and it is actively maintained, whereas Bullet and FCL are not anymore.
Coal as it currently stands unfortunately does not allow anyone to implement custom shapes, and forking it just to add our cast hull shape is not desirable (and I don't think this custom shape is a good fit for inclusion in upstream Coal as a default shape). So to be able to move forward we'd really appreciate if this PR could be considered for merging.

(I do admit this GEOM_CUSTOM solution seems somewhat clunky, but it follows from the current design of the Coal dispatch internals. I've also iterated with Coal over a [design brief](https://github.com/coal-library/coal/discussions/836) for a different solution that would do away with the enormous CollisionFunctionMatrix, ContactPatchFunctionMatrix and DistanceFunctionMatrix altogether, but this might be too invasive a change.)

test/custom_shape.cpp
```diff
@@ -0,0 +1,759 @@
...
+            &shape_, dir_local1, hint1_));
+
+    // Return the support of the convex hull of both poses (prefer pose 1 on
+    // tie).
+    support = (dir.dot(s0) > dir.dot(s1)) ? s0 : s1;
+  }
+
+  /// @brief Delegate to the underlying shape via shape_traits lookup.
+  bool needNesterovNormalizeHeuristic() const override {
+    return details::getNormalizeSupportDirection(&shape_);
```
lmontaut, line 162: I'm not sure to understand this. `details::getNormalizeSupportDirection` calls `needNesterovNormalizeHeuristic` in the `GEOM_CUSTOM` case. Isn't this an infinitely self calling situation?

include/coal/narrowphase/support_functions.h
```diff
@@ -108,6 +115,22 @@ template <int _SupportOptions = SupportOptions::NoSweptSphere,
...
+/// allowing custom shapes to participate in GJK/EPA computations.
+template <int _SupportOptions = SupportOptions::NoSweptSphere>
+void getShapeSupport(const ShapeBase* shape, const Vec3s& dir, Vec3s& support,
+                     int& hint, ShapeSupportData& support_data);
+
+/// @brief Generic ShapeBase support set function.
+/// This overload uses virtual dispatch for custom shapes.
+/// The default behavior computes a single support point.
+template <int _SupportOptions = SupportOptions::NoSweptSphere>
+void getShapeSupportSet(const ShapeBase* shape, SupportSet& support_set,
```
lmontaut: This should go in the "support set" section of the file.

include/coal/shape/geometric_shapes.h
```diff
@@ -85,6 +86,32 @@ class COAL_DLLAPI ShapeBase : public CollisionGeometry {
...
+  /// The output support point is expressed in the local frame of the shape
+  /// and does not account for the swept sphere radius.
+  /// Override getNodeType() to return GEOM_CUSTOM and this method to enable
+  /// custom shapes to participate in GJK/EPA collision and distance
+  /// computations.
+  /// @param[in] dir support direction. Always unit-length when called by Coal.
+  /// @param[out] support the computed support point.
+  /// @param[in,out] hint warm-start hint (used mainly for convex shapes).
+  /// @param[in,out] data temporary data for support computation.
+  virtual void computeShapeSupport(const Vec3s& dir, Vec3s& support, int& hint,
```
lmontaut, line 112: I am really not a fan of having the `computeShapeSupport` method in `ShapeBase`. And the fact that the default behavior is throwing, regardless of the shape.

All the shapes in coal have an implementation of the support function. This function lets the reader think that this is not the case. It should at least call `getShapeSupport` for shapes which are not `GEOM_CUSTOM`.

lmontaut:
Hi @rjoomen, sorry for the delay. The team has been super focused on the release of Pinocchio 4 so @jorisv and I haven't had the time to properly look into this PR.

First thing, we need to make sure the CI is green for this to merge, as this library is used in various environments.

Second, apart from the comments I left, I think this is a good fix for shape customization in coal for now. We can see later how to improve on that. That being said, I would really appreciate if we could wait for @jorisv input on this. There are many ways we could go about shape customization in coal and I want to make sure we are taking a maintainable road.

include/coal/narrowphase/support_functions.h
rjoomen: Done

rjoomen:
Update: CI is fixed, including the upstream ROS issues.

rjoomen:
pre-commit.ci autofix

include/coal/narrowphase/support_functions.h
```diff
@@ -63,6 +63,13 @@ namespace details {
 template <int _SupportOptions = SupportOptions::NoSweptSphere>
 Vec3s getSupport(const ShapeBase* shape, const Vec3s& dir, int& hint);
```
jorisv: Implementation of this method doesn't exists anymore

include/coal/narrowphase/minkowski_difference.h
```diff
@@ -179,6 +179,10 @@ struct COAL_DLLAPI MinkowskiDiff {
   }
 };
 
+/// @brief Runtime query for the NeedNesterovNormalizeHeuristic shape trait.
+/// Uses shape_traits for built-in shapes, virtual dispatch for GEOM_CUSTOM.
+COAL_DLLAPI bool getNormalizeSupportDirection(const ShapeBase* shape);
+
```
jorisv, line 185: You can remove it, you only use it in a test where you can directly return false.

include/coal/shape/geometric_shapes.h
```diff
@@ -86,6 +87,39 @@ class COAL_DLLAPI ShapeBase : public CollisionGeometry {
...
+  /// computations.
+  /// @param[in] dir support direction; may be non-unit-length. The support
+  /// point of a convex shape is invariant to the magnitude of dir; normalize
+  /// internally if your formula requires a unit direction (as the built-in
+  /// Cylinder does for its radial component).
+  /// @param[out] support the computed support point.
+  /// @param[in,out] hint warm-start hint (used mainly for convex shapes).
+  /// @param[in,out] data temporary data for support computation.
+  virtual void computeShapeSupport(const Vec3s& dir, Vec3s& support, int& hint,
+                                   details::ShapeSupportData& data) const;
```
jorisv, line 113: Don't add this method in `ShapeBase`, Derive from it in `CustomGeom` and add **pure virtual** methods to it.

All the template specialization and overload using `ShapeBase` should use `CustomGeom` now.

src/narrowphase/minkowski_difference.cpp
```diff
@@ -290,6 +305,9 @@ bool getNormalizeSupportDirection(const ShapeBase* shape) {
     case GEOM_CONVEX32:
       return (bool)shape_traits<ConvexBase32>::NeedNesterovNormalizeHeuristic;
       break;
+    case GEOM_CUSTOM:
+      // Use virtual dispatch for custom shapes
+      return shape->needNesterovNormalizeHeuristic();
```
jorisv, line 310: A casting to `GeomCustom`  should be done

src/narrowphase/support_functions.cpp
```diff
@@ -81,20 +81,34 @@ Vec3s getSupport(const ShapeBase* shape, const Vec3s& dir, int& hint) {
       break;
     case GEOM_PLANE:
     case GEOM_HALFSPACE:
+      support.setZero();
+      break;
+    case GEOM_CUSTOM:
+      // Use virtual dispatch for custom shapes
+      getShapeSupport<_SupportOptions>(shape, dir, support, hint, support_data);
```
jorisv, line 88: A casting to `GeomCustom` should be done.

src/narrowphase/support_functions.cpp
```diff
@@ -490,10 +521,16 @@ void getSupportSet(const ShapeBase* shape, SupportSet& support_set, int& hint,
       break;
     case GEOM_PLANE:
     case GEOM_HALFSPACE:
+      break;
+    case GEOM_CUSTOM:
+      // Use virtual dispatch for custom shapes
+      getShapeSupportSet<_SupportOptions>(shape, support_set, hint,
```
jorisv, line 527: A casting to `GeomCustom` should be done.

src/shape/geometric_shapes.cpp
```diff
@@ -57,6 +58,21 @@ using orgQhull::QhullVertexSet;
 
 namespace coal {
 
+void ShapeBase::computeShapeSupport(const Vec3s& dir, Vec3s& support, int& hint,
```
jorisv, line 61: To remove

test/custom_shape.cpp
```diff
@@ -0,0 +1,887 @@
...
```
jorisv, line 162: return false

test/custom_shape.cpp
```diff
@@ -0,0 +1,887 @@
...
+  BOOST_CHECK_EQUAL(sphere.getObjectType(), OT_GEOM);
+
+  CastSphere cast_sphere(1.0, Transform3s());
+  BOOST_CHECK_EQUAL(cast_sphere.getNodeType(), GEOM_CUSTOM);
+}
+
+/// Test that a custom sphere gives the same distance result as Coal's
+/// built-in Sphere, using both the low-level GJK API and the high-level
+/// distance() API.
+BOOST_AUTO_TEST_CASE(test_custom_sphere_vs_builtin_sphere_distance) {
```
jorisv, line 195: You are only testing the high-level api here. You can delete the GJKSolver construction that is never used.

test/custom_shape.cpp
```diff
@@ -0,0 +1,887 @@
...
+
+  Scalar radius;
+};
+
+// ============================================================================
+// CastSphere: swept volume of a sphere moving from the identity transform to
+// a given cast transform. The support function returns the point that maximises
+// dot(d, ·) over both poses (Schulman et al., 2014).
+// ============================================================================
+class CastSphere : public ShapeBase {
```
jorisv, line 107: Why two custom geometry ? The first one is already covering all

test/custom_shape.cpp
```diff
@@ -0,0 +1,887 @@
...
+  tf_box.setTranslation(Vec3s(sweep_dist + radius + Scalar(2.0), 0, 0));
+  Scalar d = distance(&cast_sphere, tf_cast, &box, tf_box, dist_req, dist_res);
+  BOOST_CHECK_GT(d, Scalar(0));
+  // Expected: gap = 2.0 - 0.5 (box half-extent) = 1.5
+  BOOST_CHECK_CLOSE(d, Scalar(1.5), Scalar(1));
+}
+
+/// Verify that getNodeType() returning GEOM_CUSTOM does NOT affect built-in
+/// shapes (they must still return their specific GEOM_* type).
+BOOST_AUTO_TEST_CASE(test_builtin_shapes_keep_their_node_type) {
```
jorisv, line 341: This test is useless

test/custom_shape.cpp
```diff
@@ -0,0 +1,887 @@
...
+    // Should not be wildly oversized (sphere conservatism is bounded by
+    // sqrt(3))
+    const Vec3s size_custom = bv_custom.max_ - bv_custom.min_;
+    const Vec3s size_builtin = bv_builtin.max_ - bv_builtin.min_;
+    BOOST_CHECK(
+        (size_custom.array() <= size_builtin.array() * Scalar(1.8)).all());
+  }
+}
+
+#ifdef COAL_HAS_OCTOMAP
```
jorisv, line 395: This only test we fill the distance/collision matrices ?
If it's the case you can remove them.

test/custom_shape.cpp
```diff
@@ -0,0 +1,887 @@
...
+      collide(&custom, Transform3s(), &octree, Transform3s(), request, result));
+  result.clear();
+  BOOST_CHECK_NO_THROW(
+      collide(&octree, Transform3s(), &custom, Transform3s(), request, result));
+}
+
+#endif  // COAL_HAS_OCTOMAP
+
+// ============================================================================
+// BVH mesh tests
```
jorisv, line 501: If it's only testing that we fill the matrix you can remove them.

test/custom_shape.cpp
```diff
@@ -0,0 +1,887 @@
...
+
+  // Symmetry: mesh vs custom should give similar distance
+  result.clear();
+  Scalar d_rev = distance(mesh.get(), tf2, &custom, tf1, request, result);
+  BOOST_CHECK_GT(d_rev, Scalar(0));
+  BOOST_CHECK_CLOSE(d, d_rev, Scalar(1));
+}
+
+/// Collision: CustomSphere vs HeightField<AABB>
+BOOST_AUTO_TEST_CASE(test_custom_shape_heightfield_collision) {
```
jorisv, line 567: If it's only testing that we fill the matrix you can remove them.

test/custom_shape.cpp
```diff
@@ -0,0 +1,887 @@
...
+  Cylinder cylinder(radius, 2.0);
+  test_pair(&custom, tf1, &cylinder, tf_near);
+
+  Ellipsoid ellipsoid(radius, radius, radius);
+  test_pair(&custom, tf1, &ellipsoid, tf_near);
+}
+
+/// Support points must be invariant to the magnitude of the support
+/// direction: GJK passes raw, possibly non-unit-length directions.
+BOOST_AUTO_TEST_CASE(test_support_direction_scale_invariance) {
```
jorisv, line 764: I don't get what is tested here

test/custom_shape.cpp
```diff
@@ -0,0 +1,887 @@
...
+      const Vec3s ss2 =
+          details::getSupport<details::SupportOptions::WithSweptSphere>(
+              shape, scale * dir, hint2);
+      BOOST_CHECK_SMALL((ss1 - ss2).norm(), Scalar(1e-12));
+    }
+  }
+}
+
+// ============================================================================
+// A GEOM_CUSTOM shape that does not override computeShapeSupport().
```
jorisv, line 808: This can be deleted with the GeomCustom subclass modification

test/custom_shape_deformed_cylinder.cpp
jorisv, line 1: custom_shape is already testing all we need.

We can maybe create an example directory to store it.

jorisv (changes requested):
@rjoomen Please **do not apply** the review right now.
I did it to well understand the architectural modification you did to support user-defined geometry.

We have to discuss other strategy with @lmontaut before.
If we approve this one we will ask you to apply the review.

### TODO

- [ ] Add a way to inherit from GeomCustom in python binding

PR #838, opened 2026-04-30
AI is taking our job (30% speed improvements)

facontidavide:
Welcome to the world of "vibe coding", where I don't even read the code.

LET ME BE CLEAR: this is a BAD pull request, one where too much work and responsibility is expected from the reviewer and the one submittting the code (ME) is not doing his homeworks.

You should be, as a minimum, offended by that.

I have no idea what these changes are, I only know that all the tests pass and that it is incredibly faster.

I tried to split the changes in individual commits that can reviewed one by one, each with a specific and self contained change.

I am sharing this hoping to find this useful, but I am not going to swear with my life that everything is 100% correct.

image


rjoomen:
Hi Davide,

Thanks for the PR and the work you and Claude put in. I think there are interesting things here, but I'll leave it up to the maintainers to respond further. 
One thing: you suggested individual small PRs, which are much easier for the reviewers to process. This, however, is all in one PR. Splitting it would be easier, I think, although more work for you.

Cheers,
Roelof

lmontaut:
Hi @facontidavide, thanks for your interest in coal.

I have to say, I am oscillating between interest and frustration while reading this PR.

Let's start by the good:
We've known for a long time that the BVH-BVH code of coal is (really) sub-optimal, and for many reasons. We haven't had the time to address these yet.

The agent found the TODO I placed some time ago in the the BVH-BVH case of coal and decided to optimise it.
In short, it got rid of the GJK solver in the BVH-BVH case and added a specialization for triangle-triangle collisions (which are called internally for BVH-BVH collisions). This is a good idea.
By simply adding this specialization, I can get the same kind of performance boost you report.
It also precomputed some stuff related to BVs. But it only did it for the BVH-BVH case which makes no sense. We need to see if this is really useful.

So, to make the scope of this PR clearer: in the case where the user doesn't care about contact information in case of collision (pure yes/no collision check), and in the case where the pair of shapes are both BVHs, the code is better optimized than before.

Let's move on to the frustration:
It takes a lot of time to extract what is good from this and make it maintainable.
The commits are all bundled together and it's hard to do a proper ablation of what made things faster/what has no impact/what might introduce a bug. 
Also, the agent added a bunch of code related to simd. This has nothing to do with the BVH-BVH benchmark and the agent simply never tested the performance impact of this.
The simd code it wrote is related to small convex-hull checks (less than ~30 points). @jorisv has already investigated this; we haven't merged it yet because it's not clear if adding it is worth it.

Anyway, thanks for sharing these ideas to improve coal. We'll try to see how we can incorporate them while keeping this software maintainable.

facontidavide:
As I said, this was a lazy first attempt. Let me split it in 8 PR, each with its own benchmark comparison

facontidavide:
closed in favor of a bunch of PR

PR #841, opened 2026-04-30
perf: devirtualize BVH traversal via templated collide/distance overloads

facontidavide:
## Summary

- Adds **templated** `collisionRecurseT<Node>` and `distanceRecurseT<Node>` next to the existing virtual-dispatch versions in `include/coal/internal/traversal_recurse.h`.
- Adds **templated** `collide<Node>` / `distance<Node>` overloads in `src/collision_node.h` that route through the templated recursers.
- Per overload-resolution rules, the templated overload outranks the existing base-pointer non-template (exact-type match beats derived-to-base conversion). Internal callers in `collision_func_matrix.cpp` and `distance_func_matrix.cpp`, which already create concrete `MeshCollisionTraversalNode<BV>` / `MeshDistanceTraversalNode<BV>` instances, automatically pick up the devirtualised path with **zero source changes at the call site**.

## Implementation notes

- `traversal_recurse.h` gains `collisionRecurseT` and `distanceRecurseT` (header-only templates) that take `Node*` instead of `CollisionTraversalNodeBase*`. The 9 virtual calls per BVH node pair (`isFirstNodeLeaf`, `isSecondNodeLeaf`, `BVDisjoints`, `leafCollides`, `firstOverSecond`, `getFirstLeftChild`, `getFirstRightChild`, `getSecondLeftChild`, `getSecondRightChild`, `canStop`) all become statically-resolved member calls on the concrete `Node` type — the inliner can fully inline them.
- `src/collision_node.h` gains `collide<Node>` / `distance<Node>` template overloads that wrap the `T`-prefixed recursers. Front-list propagation and queue-based search paths remain on the virtual codepath (those are rare and not worth the extra instantiations).
- `src/collision_node.cpp` exports `checkResultLowerBound` as `COAL_DLLAPI` so the header-only template can call it without a link-time visibility hole.

## Performance impact

**Methodology**

- Build: `cmake --build build -j12 -DCMAKE_BUILD_TYPE=Release` in isolated worktrees; separate `libcoal.so` per variant. Both base and variant compiled with stock upstream flags (no `-march=native`).
- Workload: `coal-test-benchmark` (`test/benchmark.cpp`).

| `coal-test-benchmark` total  | 62591.4             | 60417.5            | -3.47% | 15 | base 313.8 / variant 324.4 |

**Correctness gate**: full `ctest` suite passes (37/37, including the new `coal-devirt_correctness` test).

## Tests

- The cherry-picked commit already adds `test/benchmark_devirt.cpp` (a perf benchmark that calls `collisionRecurseT` directly so the templated dispatch route is exercised at build time).
- **Added** `test/devirt_correctness.cpp` — a focused correctness test that, for each `MeshCollisionTraversalNode<BV>` / `MeshDistanceTraversalNode<BV>` instantiation (`OBB`, `RSS`, `kIOS`, `OBBRSS`; OBB has no distance node), runs both the templated path (`collide(&node, ...)` / `distance(&node)`) and the virtual-dispatch path (`collide(static_cast<CollisionTraversalNodeBase*>(&node), ...)`) on identical inputs and asserts the two `CollisionResult`s / `DistanceResult`s are pairwise-equal.
- The whole point of the change is overload-resolution dispatch; a benchmark proves it routes to the templated path, but only this correctness test proves the templated path produces *exactly* the same result.

## Risk & regression analysis

- **Overload resolution silently changing**: the templated overloads outrank the non-template base-pointer overloads by C++ rules. If a future change introduced an ambiguous overload, the wrong path could be selected silently. The new correctness test catches this — if either path drifts, results disagree.
- **Front-list / queue-based variants**: not templated. Callers that opt into those still get the virtual-dispatch path. No change in behaviour.
- **ABI**: no public header signature changes. The added `COAL_DLLAPI` on `checkResultLowerBound` is a visibility expansion — safe for consumers.


lmontaut:
See #858

PR #843, opened 2026-04-30
perf: hoist GJKSolver out of mesh leaf-collision path

facontidavide:
## Summary

- Hoists the per-leaf `GJKSolver` construction in `MeshCollisionTraversalNode::leafCollides` to a `mutable GJKSolver solver_` member.
- Eliminates allocator traffic from `EPA::reset()` and the two `SimplexVertex` / `SimplexFace` `std::vector` resizes that were happening on **every** triangle-pair leaf test, not just once per query.
- The `TODO` at `traversal_node_bvhs.h:206` (pre-change) already called for this fix.
- Adds `test/benchmark_robot.cpp`: a leaf-firing benchmark using `rob.obj × rob.obj` over 2000 random transforms, P-core pinned. Makes this kind of allocator-traffic gain visible to future reviewers — the existing `coal-test-benchmark` uses extents `[-3000, 3000]` with mesh size ~10 units, so leaves never fire.

## Implementation notes

- `solver_` is `mutable` because `leafCollides` is `const` but the GJK iterations naturally mutate the solver's working state. The constructor takes the same `CollisionRequest` as before; the solver is initialised once at node construction time.
- No public API change — `MeshCollisionTraversalNode<BV>` retains the same constructor signature and member contract.
- Each leaf call now sees the same solver instance. The risk this introduces — *state pollution between leaf calls* — is locked down by the new test `coal-gjk_hoist_no_state_pollution`.

## Performance impact

The per-leaf allocation is ~3 % of total cycles only on workloads that actually fire leaves. The synthetic `coal-test-benchmark` uses transform extents that put the meshes far apart most of the time, so broad-phase culling kills the recursion before reaching leaves — synthetic Δ is therefore within noise.

**Methodology**

| `coal-test-benchmark` total  | 62457.7             | 62386.5            | -0.11% | 15 | base 239.0 / variant 563.8 |

The original `rob.obj × rob.obj` measurement (cited from the source commit, not reproducible from inside this repo's existing benchmarks alone) showed:

- `RSS deep_overlap`: 17.54 → 16.6 µs (-5.4 %)
- `RSS medium_overlap`: 20.58 → 19.65 µs (-4.5 %)
- `OBBRSS deep_overlap`: 8.88 → 8.25 µs (-7.1 %)
- `OBBRSS medium_overlap`: 9.86 → 9.27 µs (-6.0 %)

`benchmark_robot` is shipped alongside the change to make those numbers reviewer-reproducible going forward.

**Correctness gate**: full `ctest` suite passes (37/37, including the new `coal-gjk_hoist_no_state_pollution` test).

## Tests

- **Added** `test/gjk_hoist_no_state_pollution.cpp` — locks down the no-state-pollution invariant. Runs `collide()` multiple times on the *same* `MeshCollisionTraversalNodeRSS` instance with different transforms, then runs `collide()` on a *fresh* node instance with the same transforms; if the member solver had pollution between leaf calls, the two paths would diverge. Asserts collision-result equivalence (`isCollision`, `numContacts`, `distance_lower_bound`) on every transform.
- **Added** `test/benchmark_robot.cpp` — a leaf-firing benchmark for future perf comparisons (not part of the test suite; not run by `ctest`).

## Risk & regression analysis

- **State pollution across leaf calls** is the only meaningful risk introduced by the hoist. The new test exercises exactly that worst case and is the primary regression gate.
- **Constructor contract**: the solver is now built once per node (per query) instead of once per leaf. The `CollisionRequest` arrives at the same point as before, so any request-derived solver tuning still applies.
- **Public API**: unchanged. The `mutable` keyword on `solver_` is the only visible difference, and only to subclasses.

### DIFF
--- a/include/coal/internal/traversal_node_bvhs.h
+++ b/include/coal/internal/traversal_node_bvhs.h
@@ -137,7 +137,7 @@ class MeshCollisionTraversalNode : public BVHCollisionTraversalNode<BV> {
   };
 
   MeshCollisionTraversalNode(const CollisionRequest& request)
-      : BVHCollisionTraversalNode<BV>(request) {
+      : BVHCollisionTraversalNode<BV>(request), solver_(request) {
     vertices1 = NULL;
     vertices2 = NULL;
     tri_indices1 = NULL;
@@ -203,15 +203,11 @@ class MeshCollisionTraversalNode : public BVHCollisionTraversalNode<BV> {
     TriangleP tri1(P1, P2, P3);
     TriangleP tri2(Q1, Q2, Q3);
 
-    // TODO(louis): MeshCollisionTraversalNode should have its own GJKSolver.
-    GJKSolver solver(this->request);
-
     const bool compute_penetration =
         this->request.enable_contact || (this->request.security_margin < 0);
     Vec3s p1, p2, normal;
-    DistanceResult distanceResult;
     Scalar distance = internal::ShapeShapeDistance<TriangleP, TriangleP>(
-        &tri1, this->tf1, &tri2, this->tf2, &solver, compute_penetration, p1,
+        &tri1, this->tf1, &tri2, this->tf2, &solver_, compute_penetration, p1,
         p2, normal);
 
     const Scalar distToCollision = distance - this->request.security_margin;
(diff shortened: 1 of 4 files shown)


facontidavide:
pre-commit.ci autofix

lmontaut:
See 858

PR #844, opened 2026-04-30
ci: disable industrial_ci PRERELEASE

ahoarau:
Bump ros-industrial-ci to latest master and backport ros-ci from jrl-v2 (cleaned up).
The main contribution is disabling `PRERELEASE: true` which is causing issues with ros noble.

Fix for the issue referenced here https://github.com/coal-library/coal/pull/822#issuecomment-4353880072

### DIFF
--- a/.github/workflows/ros_ci.yml
+++ b/.github/workflows/ros_ci.yml
@@ -28,41 +28,51 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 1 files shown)


ahoarau:
@rjoomen That was the `PRERELEASE: true` that was causing the issue 👍

jorisv (approved):

PR #847, opened 2026-04-30
perf: prefetch BVH children before BV-pair SAT in mesh-mesh traversal

facontidavide:
## Summary

- Adds `__builtin_prefetch` hints to `MeshCollisionTraversalNode::BVDisjoints` for the four likely-next BV nodes (left + right children of both BVH sides) before running the SAT / rect-distance test.
- The SAT test takes ~50–130 ns; the prefetch resolves into L2/L3 in that window, so the recurse(c1, b2) / recurse(c2, b2) descent that follows finds the next BVNode header already warm.
- Wasted hint cost on disjoint pairs is ~4 cycles per `__builtin_prefetch`; saving when the data is needed is ~25 ns of memory latency.

## ⚠️ Open against #842 (`oriented-bv-traversal`)

This PR is conceptually downstream of [#842 (oriented BV traversal opts)](https://github.com/coal-library/coal/pull/842). On the original branch, the prefetch wraps the restructured `BVDisjoints` introduced by #842 (which uses `overlapPrecomputedRTranspose`). Cherry-picking onto pure `devel` for this PR uses the original `overlap(R, T, ...)` form on the non-identity branch — the prefetch effect is identical, only the SAT-test call differs. After #842 lands, a trivial rebase will switch this PR's non-identity branch back to `overlapPrecomputedRTranspose`.

## Implementation notes

- `include/coal/internal/traversal_node_bvhs.h`: `BVDisjoints` now binds `n1 = getBV(b1)` / `n2 = getBV(b2)` once at the top, then issues 4 `__builtin_prefetch` calls (one per child of each non-leaf node) before the SAT test. The prefetch is unconditional in the sense that GCC always emits the instruction; the CPU may treat it as a no-op on CPUs that don't support it, but the call is essentially free either way.
- The prefetch addresses are the children's `BVNode` headers — typically 64–128 bytes each. Issuing 4 prefetches before a 50–130 ns SAT call gives the L2/L3 plenty of time to fill.

## Performance impact

**Methodology**

| `coal-test-benchmark` total  | 62183.1             | 62226.4            | +0.07% | 15 | base 259.6 / variant 343.6 |

The original commit measured -1 to -2 % per joint on the `polso/gen4` mesh-mesh distance workload (a real-world workload outside this repo). The synthetic benchmark exercises the same BVH-traversal code path, so the gain should be visible there too if the BVH descent dominates the workload.

**Correctness gate**: full `ctest` suite passes — `__builtin_prefetch` is a pure performance hint with no semantic effect.

## Tests

- **No new test added** — `__builtin_prefetch` cannot in itself break correctness. The existing `coal-collision`, `coal-distance`, and `coal-distance_lower_bound` suites already exercise the BVH traversal path with strict numerical tolerances; if the refactor of `BVDisjoints` (binding `n1` / `n2` once at the top) had introduced any subtle behaviour change, those suites would catch it.

## Risk & regression analysis

- **Cache pollution**: prefetching cache lines that turn out not to be needed (when the BV pair is disjoint, no recursion happens, so the prefetched children are wasted). On modern CPUs the L2/L3 has plenty of capacity for 4 unused 64-byte lines per traversal node; the cost is bounded by the worst-case node count in a query (typically ≪ 10000).
- **Performance regression on cold-cache CPUs**: theoretically possible if a CPU has a very small L2 and the prefetched lines evict useful data. Not observed in practice on x86 / aarch64; we'd see it as a regression on `coal-test-benchmark` if it existed.
- **Compiler portability**: `__builtin_prefetch` is supported by GCC, Clang, and ICC. MSVC has `_mm_prefetch` instead — that path is left unchanged in this PR; on MSVC builds the prefetch is simply absent. The optimisation is best-effort by intent.

### DIFF
--- a/include/coal/internal/traversal_node_bvhs.h
+++ b/include/coal/internal/traversal_node_bvhs.h
@@ -151,13 +151,30 @@ class MeshCollisionTraversalNode : public BVHCollisionTraversalNode<BV> {
   bool BVDisjoints(unsigned int b1, unsigned int b2,
                    Scalar& sqrDistLowerBound) const {
     if (this->enable_statistics) this->num_bv_tests++;
+    const BVNode<BV>& n1 = this->model1->getBV(b1);
+    const BVNode<BV>& n2 = this->model2->getBV(b2);
+    // Speculative prefetch of children for both BVH sides. If this BV pair
+    // overlaps the recursion descends immediately and one of these cache
+    // lines becomes the next `getBV(c)` access. The SAT/rect-distance test
+    // below takes ~50-130 ns, plenty of time for an L2/L3 prefetch to
+    // resolve. Wasted hint cost on disjoint pairs is ~4 cycles per call,
+    // far below the saving when the data is needed.
+    if (!n1.isLeaf()) {
+      const int c = n1.first_child;
+      __builtin_prefetch(&this->model1->getBV(c));
+      __builtin_prefetch(&this->model1->getBV(c + 1));
+    }
+    if (!n2.isLeaf()) {
+      const int c = n2.first_child;
+      __builtin_prefetch(&this->model2->getBV(c));
+      __builtin_prefetch(&this->model2->getBV(c + 1));
+    }
+
     bool disjoint;
     if (RTIsIdentity)
-      disjoint = !this->model1->getBV(b1).overlap(
-          this->model2->getBV(b2), this->request, sqrDistLowerBound);
+      disjoint = !n1.overlap(n2, this->request, sqrDistLowerBound);
     else {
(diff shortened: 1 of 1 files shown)


facontidavide:
Closing this one. The synthetic `coal-test-benchmark` shows +0.07 % (variant slightly slower, well within the ~0.5 % stdev), and the original real-mesh measurement on the upstream branch was only -1 to -2 % per joint. The cost/benefit ratio is the worst in the stack and the change isn't load-bearing for any of the other PRs in the series.

Deferring until a profiling pass identifies the BVH-header fetch as a bottleneck on a specific workload that justifies it. The branch `perf/bvh-prefetch` on the fork (https://github.com/facontidavide/coal/tree/perf/bvh-prefetch) is preserved for easy revival.

PR #744, opened 2025-08-12
Remove useless boost components direct dependency

jorisv:
Some boost components had been added as a direct coal or coal binding dependency. Since these components (system, thread, date_time, chrono) are not directly used, we should leave it to the Boost CMake module.

Fix #743 

We can't test it with boost 1.89 since it's not available in conda-forge. But I think this version will be available soon.
@cho-m can you test it locally ?

PR #848, opened 2026-04-30
build: fix Boost dependency discovery on Ubuntu 26.04

mjcarroll:
I've explicitly added `system` to the Boost requirements to ensure consistency across platforms.

Could we please get a new release of `coal` to unblock binary builds for the ROS 2 Lyrical distribution?

Reference build: https://build.ros2.org/view/Lbin_uR64/job/Lbin_uR64__coal__ubuntu_resolute_amd64__binary/2/


nim65s:
I think this is already fixed by https://github.com/coal-library/coal/pull/744 and this PR would revert the fix.

And I guess you are right: we need a release.

mjcarroll:
Thanks! In that case, just a new release would be great as it unblocks things like MoveIt further up the ROS stack.

nim65s:
we'll work on it :)

Issue #837, opened 2026-04-29
30% speedup. What is the best way to contribute?


facontidavide:
Hi,

I just wanted to contact you first, to make sure that I can contribute in the best way possible to this project.

In the last 24 hours I had AI optimizing Coal with a good level of success. On many practical mesh-to-mesh scenarios I got an average 30% speed improvement.

What is the best way to submit these changes as PR? I was thinking to default to small PRs with individual changes and regression tests when appropriate, but I prefer to double check first.

Also, how do you feel about AI generated code?

nim65s:
Hi @facontidavide 

Yes, "small PRs with individual changes and regression tests when appropriate" is perfect, thanks !

Our AI policy is not out yet, but if you can vouch for what the AI say that will be fine I guess

facontidavide:
I am assuming the unit tests are a sufficient guardrail. About "vouching"... Man, I have no frickin idea what this code does 🫠

facontidavide:
If you want to merge only one, this is the winner: #842

lmontaut:
Hi @facontidavide, thanks for your feedback on BVH-BVH collisions!
I opened #858 and I notably took part of your #842 PR to improve performances.

PR #858, opened 2026-05-13
Improve the performances of BVH-BVH collision detection

### LINKED ISSUE #857: Triangle-mesh broadphase ~1.8× slower than FCL 0.6.1; precision and linkage ruled out`
## Summary

Evaluating a migration from FCL 0.6.1 → Coal 3.0.2 for triangle-mesh collision detection in a robotics codebase. On a synthetic scene, Coal's broadphase self-collide is consistently ~1.8× slower than FCL across every build configuration I tried (default, float-precision, static linkage). Distance queries are ~1.08× slower, single-pair narrowphase is ~1.14× slower.

Results are numerically identical between the two libraries (contact counts and distance sums match across 2000-iteration runs), so this is purely a perf observation, not a correctness one.

Posting in case the workload is useful, and because the patches I had to apply to enable float precision look like real upstream bugs worth flagging separately.

## Workload

Synthetic scene approximating 

[fcl_vs_coal_benchmark_test.cpp](https://github.com/user-attachments/files/27611144/fcl_vs_coal_benchmark_test.cpp)

:

- 1 static pedestal cylinder (24-segment tessellation)
- 1 static pallet box
- 12 static stacked product boxes (4×3 grid, 0.28 m cubes)
- 6 dynamic robot-link cylinders (16-segment tessellation, transforms swept each iteration)
- 1 dynamic gripper box

Total: 21 triangle-mesh BVHs over `OBBRSS`. No convex primitives — all geometry is mesh, so none of the GJK/EPA improvements apply.

Three workloads, 2000 iterations each, with a deterministic per-iteration transform sweep for the 7 dynamic objects:

1. **Broadphase self-collide**: `manager.collide(callback)`
2. **Pairwise distance**: gripper vs each of 14 static objects, summed
3. **Single-pair narrowphase collide**: gripper vs pallet (10 000 iters)

## Setup

- Linux x86_64, GCC 11, `-O3 -DNDEBUG`, single-threaded
- Eigen 3.4.0
- FCL 0.6.1 (Conan Center recipe — static archive, `BVHModel<OBBRSS<float>>`, `DynamicAABBTreeCollisionManager<float>`)
- Coal 3.0.2 (Conan Center recipe initially; later patched local recipe for variants)
- Timing via `std::chrono::steady_clock`, 5-iter warm-up before measuring

## Results

Three runs per configuration, stable across runs. Ratios relative to FCL baseline:

| Config | Broadphase | Distance | Single-pair |
|---|---|---|---|
| **FCL 0.6.1** (float, static) — baseline | 1.0× | 1.0× | 1.0× |
| Coal 3.0.2 (shared, double — Conan default) | **1.71× slower** | 1.13× slower | ~1.0× |
| Coal 3.0.2 (shared, float — patched) | **1.82× slower** | 1.08× slower | 1.17× slower |
| Coal 3.0.2 (static, float — patched) | **1.87× slower** | 1.08× slower | 1.14× slower |

Absolute numbers for the static + float Coal config:

- FCL broadphase: ~10.7 µs/iter
- Coal broadphase: ~20.0 µs/iter
- FCL distance (14 pair queries): ~306 µs/iter
- Coal distance: ~330 µs/iter

### Precision (refuted)

Hypothesized that Coal's default double precision was the cost vs FCL's float. The `COAL_USE_FLOAT_PRECISION` CMake flag is recognized in `src/CMakeLists.txt` but is a no-op at v3.0.2 — `include/coal/data_types.h` hard-codes `typedef double CoalScalar` and `typedef double FCL_REAL` with no `#ifdef`. The actual float support only exists on the `devel` branch.

Patched the released source in-place to test the hypothesis anyway:

- `include/coal/data_types.h`: `typedef double FCL_REAL` → `float`, `typedef double CoalScalar` → `float`. All `Vec3s` / `Matrix3s` / `Transform3s` derive from these so the change propagates.
- `include/coal/narrowphase/narrowphase.h` (2 sites) and `src/narrowphase/gjk.cpp` (1 site): bare `0.` double literals passed to `std::max` / `std::min` alongside `CoalScalar` args, promoted to `CoalScalar(0)`.
- `include/coal/shape/geometric_shapes.h` (3 sites) and `src/shape/geometric_shapes.cpp` (1 site): `ConvexBase::offsets` is declared as `std::shared_ptr<std::vector<double>>` rather than `std::vector<CoalScalar>`. Patched to `std::vector<CoalScalar>`.

Library builds cleanly after the patch and results still match FCL exactly. Result: broadphase ratio went from 1.71× → 1.82× (no improvement, within noise); distance ~5% better. **Precision is not the cost.**

### Static vs shared linkage (refuted)

Hypothesized PLT/GOT indirection on each Coal call vs inline-able FCL static-archive calls. The Conan recipe declares `package_type = "shared-library"` and `src/CMakeLists.txt:139` hard-codes `add_library(${LIBRARY_NAME} SHARED ...)` — neither honors `BUILD_SHARED_LIBS`. Patched both: removed the explicit `SHARED` keyword, switched the recipe to `package_type = "library"` with a `shared` option, passed `CMAKE_POSITION_INDEPENDENT_CODE=ON`. Built `libcoal.a`, verified static linkage via `ldd`.

Result: no measurable change in any of the three workloads. **Linkage is not the cost either.**

## Where the cost might be (speculation, not investigated)

The consistent ~1.8× factor across all configurations suggests the gap is in the broadphase code itself. Plausible candidates I didn't investigate:

- **Callback dispatch:** Coal's broadphase callback is `CollisionCallBackBase::collide()`, a virtual method on a struct. FCL uses a function pointer. At thousands of candidate pairs per broadphase call, vtable dispatch could be material.
- **Per-object state size:** Coal's `CollisionObject` carries additional fields (safety margins, GJK variants enum, etc.) which may bloat the broadphase's working set.
- **Algorithmic differences:** Different tree-balancing thresholds, candidate-pair pruning, or AABB-update strategies in the dynamic AABB tree manager.

Happy to share a profile if there's interest.

## Notes for maintainers

Two things from the float-precision work that look upstream-worthy regardless of any Coal-vs-FCL comparison:

1. **`ConvexBase::offsets` is `std::vector<double>` instead of `std::vector<CoalScalar>`** in `geometric_shapes.h` and `geometric_shapes.cpp`. This unconditionally breaks any float-precision build (including the one on `devel`, presumably). Looks like a missed substitution.
2. **`src/CMakeLists.txt:139` hard-codes `SHARED`** in the `add_library` call, so `BUILD_SHARED_LIBS=OFF` is silently ignored. Worth honoring the standard CMake switch.

## Reproduction

The benchmark is a single self-contained ~350-line GTest file — no external mesh assets, deterministic transforms, prints timings to stdout. Happy to share the source if useful for verification.

lmontaut:
Multiple users have recently complained about the performance of BVH-BVH collision detection #857 #837.
This PR impacts BVH-BVH collisions:
- removes virtualisation for traversal nodes (similar to #841)
- precomputes relative transform (similar to #842)
- switch to triangle-triangle specialization instead of the iterative GJK+EPA algorithm
- adds a fast path if `request.enable_contact` is `false` and `request.security_margin = 0` (nice for people doing motion planning)

I also added a test to make sure the GJK and triangle-triangle specializations are identical.

No API change

This should make BVH-BVH about 2x faster.

closes #857


include/coal/internal/traversal_node_bvhs.h
```diff
@@ -200,19 +200,105 @@ class MeshCollisionTraversalNode : public BVHCollisionTraversalNode<BV> {
...
+        (this->request.security_margin == 0)) {
+      // Fast path: no security margin, we don't care about contact or tight
+      // lower bound so there is no need for more expensive distance
+      // computation.
+
+      // Transform Q1/Q2/Q3 into mesh1's local frame using the pre-computed
+      // relative transform RT = (R1^T*R2, R1^T*(t2-t1)).
+      // P1/P2/P3 stay in mesh1's local frame — zero transform cost.
+      Vec3s Q1r, Q2r, Q3r;
+      if (RTIsIdentity) {
```
sgayda2, line 213: sorry about the drive by comment, i just wanted to ask.

I believe that one of the tweaks from the original PR set was to make RTIsIdentity a true compile time constant and let the compiler strip the if/else accordingly. It might still be worth doing this type of change (not strictly required in this PR). Or did you find that it was not worth it?

lmontaut: Hey @sgayda2 what PR are you refering to?

I'd love to use consexpr but coal is <= c++17. I just tried on my local benchmarks with constexpr and I see almost no gain. I think we can wait for the next coal release to maybe bump to C++17 and use constexpr everywhere.

lmontaut:
Note: #846 proposed to use a different algo than the one used in this PR for triangle-triangle intersection (for now we use SAT).
I tried it, and it's a little bit better but it does not deal with coplanar triangles + it's harder to maintain. If someone wants to use the Möller algo from #846 but deal with its edge cases, you are more than welcome.

include/coal/internal/traversal_node_bvhs.h
sgayda2: Thats my bad, i guess i imagined it. I was thinking about #838 and its subsequent split PRs. Its also possible the compiler is smart enough to recognize it as constant and do the optimization anyway so there wouldn't be any difference in the generated code.

As for coal being <= c++17 i did not pay enough attention to that and did not realize this was the case. Definitely moving this to 20+ would be great, but it depends on the users/deps.

facontidavide:
Even if you closed all my PRs, I hope they provided some source of inspiration 😁

Hopefully I can get a small acknowledgement in the release note/ changeling?

jcarpent:
@lmontaut @jorisv Could we merge this PR?

jcarpent (approved):

include/coal/narrowphase/narrowphase.h
```diff
@@ -358,6 +358,73 @@ struct COAL_DLLAPI GJKSolver {
...
+      Scalar depth1((t1.a - t2.a).dot(normal));
+      Scalar depth2((t1.a - t2.b).dot(normal));
+      Scalar depth3((t1.a - t2.c).dot(normal));
+      distance = -std::max(depth1, std::max(depth2, depth3));
+    } else {
+      // No collision
+      // TODO On degenerated case, the closest point may be wrong
+      // (i.e. an object face normal is colinear to gjk.ray
+      // assert (dist == (w0 - w1).norm());
+      assert(this->gjk.ray.norm() > this->gjk.getTolerance());
```
jorisv, line 422:
```suggestion
      COAL_ASSERT(this->gjk.ray.norm() > this->gjk.getTolerance());
```

Is it a good idea to cal COAL_LOG_WARNING ?

include/coal/narrowphase/narrowphase.h
```diff
@@ -358,6 +358,73 @@ struct COAL_DLLAPI GJKSolver {
...
+    normal = normal_.cast<Scalar>();
+    Scalar distance = Scalar(this->gjk.distance);
+
+    if (gjk_status == details::GJK::Collision) {
+      Vec3s u((t1.b - t1.a).cross(t1.c - t1.a));
+      normal = u.normalized();
+      Scalar depth1((t1.a - t2.a).dot(normal));
+      Scalar depth2((t1.a - t2.b).dot(normal));
+      Scalar depth3((t1.a - t2.c).dot(normal));
+      distance = -std::max(depth1, std::max(depth2, depth3));
```
jorisv, line 416: TODO @jorisv : Check if we can simd that wit Eigen::Array

include/coal/narrowphase/narrowphase.h
```diff
@@ -358,6 +358,73 @@ struct COAL_DLLAPI GJKSolver {
     return distance;
   }
 
+  /// @brief Specialization of `shapeDistance` for the case where both shapes
+  /// are triangles. It is more efficient to pre-compute the relative
+  /// transformation between the two shapes before calling GJK/EPA.
+  Scalar shapeDistance(const TriangleP& s1, const Transform3s& tf1,
```
jorisv, line 364: What's the difference between this implementation and the one in `distance/triangle_triangle.cpp` ?

It feel weird to me me do a specialization here, where it should be handled by the distance_matrix

src/collision_node.cpp
```diff
@@ -35,56 +35,5 @@
...
-    distanceRecurse(node, 0, 0, front_list);
-  else
-    distanceQueueRecurse(node, 0, 0, front_list, qsize);
-
-  node->postprocess();
-}
-
-}  // namespace coal
+// collide() and distance() are now inline templates in collision_node.h.
+// This file is intentionally empty.
```
jorisv, line 39: Why don't remove it ? It's removed from the CMakeLists.txt

src/traversal/traversal_recurse.cpp
```diff
@@ -35,326 +35,6 @@
...
-
-  for (front_iter = append.begin(); front_iter != append.end(); ++front_iter) {
-    front_list->push_back(*front_iter);
-  }
-}
-
-}  // namespace coal
+// All traversal recurse functions are now inline templates in
+// include/coal/internal/traversal_recurse-inl.h.
+// This file is intentionally empty.
```
jorisv, line 40: Why keeping it ? It's not compiled anymore.

include/coal/internal/intersect.hxx
```diff
@@ -325,6 +325,186 @@ Project<Scalar>::projectTetrahedraOrigin(const Vec3& a, const Vec3& b,
...
+                     const Vec3& q3) {
+  using Scalar = typename Vec3::Scalar;
+  const Scalar P1 = ax.dot(p1), P2 = ax.dot(p2), P3 = ax.dot(p3);
+  const Scalar Q1 = ax.dot(q1), Q2 = ax.dot(q2), Q3 = ax.dot(q3);
+  const Scalar mn1 = std::min(P1, std::min(P2, P3));
+  const Scalar mx2 = std::max(Q1, std::max(Q2, Q3));
+  if (mn1 > mx2) return false;
+  const Scalar mx1 = std::max(P1, std::max(P2, P3));
+  const Scalar mn2 = std::min(Q1, std::min(Q2, Q3));
+  return mn2 <= mx1;
```
jorisv, line 345: TODO @jorisv: Check about SIMD and Eiger::Array

include/coal/internal/traversal_node_bvhs.h
jorisv: @lmontaut C++ 17 will be minimal version of the next release. So you can use constexpr.

jorisv (changes requested):

jorisv:
@lmontaut You speak about devirtualization, but as far as I understand, it's more about inlining function or did I miss something ?

Issue #654, opened 2025-02-05
C


esvaran-eureka:
[Height_Field_Collision_Checking_Final.zip](https://github.com/user-attachments/files/18668020/Height_Field_Collision_Checking_Final.zip)

The above folder contains 'height_field_script.py', which serves as a standalone script that reproduces the issue we're encountering currently in our environment. So, our task is to check for valid gripper positions for picking up objects in a given surrounding. Currently, the mechanism we have in place detects collisions between the gripper and the surrounding environment, which we're depicting as a point cloud using hppfcl.makeOctree() and the ComputeCollision function. The gripper, in this scenario, consists of three distinct objects (it's a two-finger gripper, so the base and the two fingers) positioned at different locations in the environment, depending on where the objects are placed. Now, these objects are placed inside of a bin, so grasp positions under the bin edges, despite not coming in contact with the point cloud itself are considered **valid, whereas they should be invalid**. To correct this behavior, we attempted to create a height field from the point cloud following this [script](http://docs.ros.org/en/melodic/api/pinocchio/html/collision-with-point-clouds_8py_source.html).

I can confidently verify that the point cloud and height field are overlapping in my representation as I've visually confirmed it. I've included the code for visualizing it with plotly as well, just in case. As far as I'm aware, the collision checking mechanism with HeightFieldOBBRSS doesn't seem to work as intended, as I'm getting quite an unexpected result for the valid gripper positions. However, I will delve into the math slightly to explain the approach I took.

The point cloud is constructed from a top-down view with the translation matrix [0, 0.4, 1.5], so z-values increase towards the ground. The height field construction won't work from this reference frame as the z-axis is inverted, so I simply altered the z-values using a simple transformation, so now I'm operating from the reference frame [0, 0.4, 0] with the normal z-axis with z-values increasing upwards, which is what you'll see in the heights array and the transform that I've set. I can verify that collision checking does work because if I replace the heights array with a plane that exists above all the gripper positions, it rejects everything, so I'm quite unsure as to why it doesn't work in this scenario, especially considering that the point cloud and height field are overlapping (collision checking with the point cloud **does work**).

Here's a picture of what it considers valid in our environment, but that position is most certainly underneath the bin edge, whereas a few positions (most notably the bottommost position on the central object) definitely aren't in contact with any object or the height field.

![Image](https://github.com/user-attachments/assets/7b54fabb-b888-4d5f-b2f7-0cc88151f8ec)

[...]

PR #879, opened 2026-07-22
Fix hfield transform

conlain-k:
# Fix heightfield contact correction for transformed heightfields

First, thank you to the developers for your hard work on Coal.

I noticed an issue in the witness-point calculations for heightfields and believe I have narrowed down a fix. This may also be related to [#654](https://github.com/coal-library/coal/issues/654).

## Problem

The heightfield-specific `binCorrection()` path mixed heightfield-local bin geometry with world-space contact and support points. Existing identity-pose tests did not expose the mismatch. I originally encountered this on version 3.0.1 and reproduced it on the `devel` branch as well.

For example, colliding a sphere near the center of a translated flat heightfield returned a witness in its local coordinates, rather than world coordinates:

```text
actual:   [0.09, -0.11, 0]
expected: [2.05, -2.07, 0]
```

In this failure path the penetration depth can still be correct, which made the problem easy to miss. Rotated heightfields could also return an incorrectly transformed normal.

## Reproduction

The new test case in `hfields.cpp` compares a flat `HeightField<OBBRSS>` with an analytically equivalent box under identity, translated, and translated-and-rotated poses. It checks penetration, both witnesses, and the contact normal using the default `CollisionRequest` settings.

The test fails on `devel` before this change and passes with the fix applied.

## Fix

This change performs face selection, support mapping, and projection in the heightfield-local frame. It then transforms the corrected witnesses and normal back to world space. The public heightfield API is unchanged.

### DIFF
--- a/include/coal/internal/traversal_node_hfield_shape.h
+++ b/include/coal/internal/traversal_node_hfield_shape.h
@@ -383,8 +383,8 @@ bool binCorrection(const ConvexTpl<Polygone>& convex,
     // sphere swept radius of the shape.
     // TODO: take into account the swept-sphere radius of the bin.
     const Vec3s _support = getSupport<details::SupportOptions::WithSweptSphere>(
-        &shape,
-        -local_shape_pose.rotation().transpose() * local_face_normal, hint);
+        &shape, -local_shape_pose.rotation().transpose() * local_face_normal,
+        hint);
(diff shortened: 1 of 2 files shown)


jorisv:
Hello @conlain-k,

Thanks for the contribution. Right now, we don't have the time to review it but we will come back to you around September.

PR #884, opened 2026-08-18
Fix ContactPatchResult::swapObjects normal and point indexing

### LINKED ISSUE #883: Title ContactPatchResult::swapObjects flips patch frame but does not update direction
Body Summary When ComputeContactPatch swaps geometries it calls ContactPatchResult::swapObjects(), which reflects the ContactPatch frame (flips x and z axes). swapObjects() does not update ContactPatch::direction, so ContactPatch::getNormal() (which returns +/- tf.col(2) depending on direction) can end up with the wrong sign compared to the original Contact.normal. This manifests as patch.getNormal() being inverted relative to the collision contact normal (e.g. collision normal (0,0,-1) but contactPatch.getNormal() returns (0,0,1)).
I also noticed a bug in the point-flip loop: the implementation uses patch.point(i) where it should use patch.point(j), so only one point is touched repeatedly instead of all points.
suggested replacement for swapObjects:
` for (size_t j = 0; j < patch.size(); ++j) {
      patch.point(j)(0) *= Scalar(-1);  // only invert the x-axis
    }`
`  // Toggle the patch direction so getNormal() keeps reporting the normal
    // pointing from shape1 to shape2 for the original ordering.
    patch.direction =
        (patch.direction == ContactPatch::PatchDirection::DEFAULT)
            ? ContactPatch::PatchDirection::INVERTED
            : ContactPatch::PatchDirection::DEFAULT;`

rjoomen:
Fixes #883.

## The defects

`ContactPatchResult::swapObjects` mirrors each patch frame (negating columns 0 and 2 of the rotation) but does not flip `ContactPatch::direction`, so `ContactPatch::getNormal` returns the negation of the contact normal the patch was built from. Its point loop also reads `patch.point(i)` — the *patch* index — inside a loop over `j`, so it reflects one point of each patch `patch.size()` times and leaves the others alone.

Measured on a box against a 2-triangle `BVHModel<OBBRSS>` plate, before the fix:

| | shape first (`swapObjects` arm) | mesh first (control) |
| --- | --- | --- |
| `contact.normal` | `(0, 0, -1)` | `(0, 0, +1)` |
| `patch.getNormal()` | `(0, 0, +1)` ❌ | `(0, 0, +1)` ✔ |

The point-loop half is currently inert through the public API: every cell of the contact-patch matrix on the `[BVH][GEOM]` and `[HFIELD][GEOM]` half routes to `BVHShapeComputeContactPatch`, `HeightFieldShapeComputeContactPatch` or `BVHComputeContactPatch`, all of which emit a one-point patch — and with one point, `point(i)` clamps to `point(0)` and is reflected exactly once, which is the intended result. It is reachable from a manually built or Python-constructed `ContactPatchResult`, and would become live the moment those arms gain a real multi-point patch.

## The fix

Flip `direction` alongside the frame, and reflect the points by iterating `patch.points()` so the loop cannot be mis-indexed again. This makes `swapObjects` a pure change of representation: the world position of every point and the normal `getNormal` reports are invariant under it.

## Tests

- `box_mesh_swapped_geometries` — box-first against a plate mesh, with the mesh-first order as the control, asserting the patch agrees with its contact in both orders.
- `swap_objects` — two three-point patches through `swapObjects` directly, covering the point loop that the one-point dispatch arms cannot reach.

Both fail on `devel` (7 assertions), pass with the fix; the full suite is 36/36 green.

## Two questions for a maintainer

1. **Is the correction needed at all on this arm?** The patch functions the swapping arm dispatches to discard `o1`/`tf1`/`o2`/`tf2` (five `COAL_UNUSED_VARIABLE` lines) and build every frame from the passed-in `CollisionResult`, whose contacts are already in the caller's order. So the patches are in the caller's convention before `swapObjects` runs, and deleting the two `result.swapObjects()` calls in `src/contact_patch.cpp` would make the shape-first result identical to the mesh-first one rather than merely equivalent under `getNormal`. I kept the call and fixed the function, because `swapObjects` is public API and is wrong as it stands either way — but if you would rather drop the calls, say so and I will.

   The flip side, and the reason this choice is not neutral: flipping `direction` is correct for any producer that takes its normal from the passed-in `CollisionResult` (all of them today). A future patch function that instead honoured its swapped arguments and emitted an o2→o1 normal would need the old flip-only behaviour. Whichever convention you prefer, it would be worth stating in the `computeContactPatch` docs.

2. **`INVERTED` becomes user-visible.** `PatchDirection`'s doc said the inverted type "should only be used for internal Coal computations"; before this fix every patch handed to a caller was `DEFAULT`. I amended that doc rather than leave it contradicting the code. Note the consequence: `ContactPatch::isSame` and `operator==` compare `direction`, so the shape-first and mesh-first patches of the same collision — same normal, same world points, same penetration depth — still compare unequal.

Separately, and not folded in here: `distance()`'s shape-first correction (`src/distance.cpp:86-89`, duplicated at `:147-150`) hand-rolls the swap and omits `b1`/`b2`, where `collide()` calls the complete `CollisionResult::swapObjects()`. Happy to open a separate issue for it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rjoomen:
Nix CI failure seems an unrelated termination of the runner.

lmontaut (approved):

PR #885, opened 2026-08-24
Fix specialized shape to shape min distance update and enable_signed_distance flag 

mjd3:
Fixes two issues:
- The SHAPE_SHAPE_DISTANCE_SPECIALIZATION / SHAPE_SELF_DISTANCE_SPECIALIZATION wrappers assigned `result.min_distance` (and o1/o2/nearest_points/normal) unconditionally, while the generic ShapeShapeDistancer path checks `request.isSatisfied` and merges through `DistanceResult::update`. When a DistanceResult is shared across several pairs (e.g., broadphase checks) a specialized pair evaluated after a closer pair overwrote the minimum, so the query returned the last evaluated pair instead of the closest one. Fix is to merge with `DistanceResult::update`, like the generic path does; added regression test for this behavior as well.
- Analytic distance functions in src/distance ignore the `compute_signed_distance` argument, so with `DistanceRequest::enable_signed_distance = false` a specialized pair could still report a negative min_distance (different from documentation/GJK path). Fix is to clamp the result at zero; the witness points and normal are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - Boost::thread
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


mjd3:
@jorisv or @nim65s let me know if there's anything else you need for this PR; would love to get this PR in eventually since it fixes a pretty major issue with broadphase distance checks in my current project.

PR #886, opened 2026-08-25
Replace undefined pointer arithmetic in octree b1/b2 with a node handle

rjoomen:
## The problem

The octree side of a contact reported `b1`/`b2` as

```cpp
Contact(tree1, c.o2, static_cast<int>(root1 - tree1->getRoot()), c.b2, ...)
```

`root1` and the tree root are separate heap allocations, not elements of one array, so this pointer subtraction is undefined behaviour. AddressSanitizer's invalid-pointer-pair check reports it directly on real octomap data:

```
==22==ERROR: AddressSanitizer: invalid-pointer-pair: 0x502000000470 0x502000000010
    #0 in main probe.cpp:13
0x502000000470 is located 0 bytes inside of 16-byte region [0x502000000470,0x502000000480)
allocated by thread T0 here:
    #0 operator new(unsigned long)
    #1 (/lib/x86_64-linux-gnu/liboctomap.so.1.9+0x19058)
```

(built with `-fsanitize=address,pointer-subtract`, run with `ASAN_OPTIONS=detect_invalid_pointer_pairs=2`).

Two properties make the result unusable beyond the formal problem. Pointer subtraction is **scaled** — `p - q` is `(byte difference) / sizeof(T)` — so it is an element count for an array these nodes were never in. And it is **signed**: a node below the root gives a negative value, and one exactly a node-width below gives `-1`, which is indistinguishable from the sentinel meaning "no index". Consumers reading `b1` for an octree contact got a value that changed between runs of the same binary on the same data.

## Why a handle, and not an index

This field carries node *identity* on the octree path, and always has. Until the result types were unified onto the generic `Contact`, octree traversals returned a dedicated struct holding the node itself:

```cpp
struct OcTreeShapeCollisionPair
{
  /** \brief The pointer to the octree node */
  const OcTree::OcTreeNode* node;
  ...
};
```

`c90f59ef` (2012-07-25) replaced that with `Contact`, whose `b1`/`b2` are `int`. A pointer no longer fit, and `root1 - tree1->getRoot()` is how it was made to fit. So the value has never been a cell index — the `b1`/`b2` documentation claiming "the id of the cell" has been aspirational since then — and callers that group contacts per cell depend on the identity it does carry.

`coal::octreeNodeHandle` keeps that identity and makes it well defined:

- `reinterpret_cast<std::uintptr_t>` is implementation-defined, not undefined;
- dividing by `alignof` rather than `sizeof` makes the mapping injective by rule, since every address is a multiple of its alignment — the old expression was injective only because the allocator happens to over-align 16-byte nodes;
- clearing the sign bit keeps the handle disjoint from the no-index sentinel, so the `-1` case above cannot recur.

Two nodes collide only if they lie 16 GiB apart in memory. On the test octree all 1385421 occupied leaves receive distinct handles.

The handle is documented as opaque: not a cell index, no spatial or enumeration meaning, and not reproducible across processes. Use the contact position if you need to know which cell was touched.

## Changes

- All 12 sites report `octreeNodeHandle` where the pointer difference was, across the shape, mesh, octree and height field traversals and their distance variants. The other operand's real primitive index — a mesh triangle id — is untouched, as is the sentinel on the height field side.
- `b1`/`b2` documentation in `collision_data.h` updated on all four fields.
- `octree_reports_stable_node_handle` in `test/octree.cpp` covers every traversal that can produce an octree contact: handles are never negative, and so never the sentinel, repeating a query reproduces them exactly, and they are distinct per node where contacts are in bijection with nodes. A mesh operand keeps its triangle id and each `(node, triangle)` pair is reported once. Both operand orders are exercised for the height field, since the handle has to follow the octree across the swap.
- The same case covers the distance path, which reaches these sites through `DistanceResult::update`: the octree against a shape, against a mesh, and against itself. The shape side reports the sentinel and the mesh side its triangle id. Height fields are absent there because octree/height-field distance is not implemented.

## Verification

Full suite green (36/36). The new test fails against the previous expression, as it should.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - If the `o1/o2` pointers are different, we check whether or not the underlying geometries are the same. This is typically important in the context of serialization.
   - Fix using NaN to initialize collision data (`Contact`, `CollisionResult`, `DistanceResult`). This prevents the absurd `Contact contact; contact == contact; // false` problem.
   - Fix NaNs coming from GJK/EPA when the algorithms (correctly) early stopped. NaNs indicate failure. In the case that GJK/EPA early stopped but ran fine, we set non-computed data to inf instead of NaN.
+- Octree contacts no longer report an undefined-behaviour `b1`/`b2`. The value was computed as `node - tree->getRoot()`, a pointer subtraction between two unrelated heap allocations, so it was undefined behaviour and could be an arbitrary signed offset. Octree contacts and distance results now report `coal::octreeNodeHandle`, a well-defined non-negative handle with the same meaning the field has carried since the node pointer was folded into `Contact`: unique per node and stable while the node lives, but not a cell index ([#886](https://github.com/coal-library/coal/pull/886))
 
 ### Changed
 - Float precision ([#665](https://github.com/coal-library/coal/pull/665))
--- a/include/coal/collision_data.h
+++ b/include/coal/collision_data.h
@@ -66,13 +66,15 @@ struct COAL_DLLAPI Contact {
   /// @brief contact primitive in object 1
   /// if object 1 is mesh or point cloud, it is the triangle or point id
   /// if object 1 is geometry shape, it is NONE (-1),
-  /// if object 1 is octree, it is the id of the cell
+  /// if object 1 is octree, it is a node handle, see
+  /// @ref coal::octreeNodeHandle
   int b1;
 
   /// @brief contact primitive in object 2
   /// if object 2 is mesh or point cloud, it is the triangle or point id
   /// if object 2 is geometry shape, it is NONE (-1),
-  /// if object 2 is octree, it is the id of the cell
+  /// if object 2 is octree, it is a node handle, see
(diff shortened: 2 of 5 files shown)


rjoomen:
Force-pushed: the branch previously reported `NONE` for the octree side. Digging into the history showed that `b1` carries node *identity* there rather than an index — it held the `OcTreeNode*` itself until the result types were unified onto `Contact` — so discarding it would have broken callers that group contacts per cell. It now reports a well-defined opaque handle instead, which removes the undefined behaviour without losing the information. Description updated accordingly.

rjoomen:
The red `nix` jobs are unrelated to this PR. They are killed with exit 143 while building the `python3.13-eigenpy` dependency, before coal itself is reached, and they have failed the same way on every push to this branch — including the earlier revision with different contents. The same failure hit `fix/contact-patch-swap-direction` on 2026-08-18, and `devel` has not run this workflow since 2026-07-07.

This change touches only `octree.h`, `collision_data.h`, `traversal_node_octree.h` and `test/octree.cpp`. The ROS, Windows, macOS, pre-commit and changelog checks are green.

nim65s:
Yes, CI cache got cold, I'll fix this tomorrow, but we have plenty of other CIs so this will be ignored for now

PR #895, opened 2026-09-22
Fix misuse of `static` in the getShapeSupport(Box) method

### LINKED ISSUE #894: [Bug]: Box support function uses an erroneous conditional function-local static variable
### 🐞 Bug

The support function (part of GJK and EPA) for [boxes](https://github.com/coal-library/coal/blob/devel/src/narrowphase/support_functions.cpp#L142-L145), like that for `Cone` and `Cylinder` uses an `inflate` factor. `Box` however, is unique in that the inflation value depends on the sampling direction.

The problem is that the `Box`'s support function stores the inflation as a `static const` variable. So, how things get inflated depends solely on the orientation of `dir` the very first time the method is invoked. Subsequent invocations, regardless of their `dir` values will pick up this first invocation, even if the logic would imply the opposite.

Ultimately, given a set of queries (at its simplest, a set of two queries), the distances reported will (and even the colliding state could) change based on the *order* of evaluation.

### Reproduction steps

```python
#!/usr/bin/env python3
"""Illustrates an order-dependence bug in coal's box support function.

`getShapeSupport(const Box*, ...)` in src/narrowphase/support_functions.cpp
picks its edge-case tie-breaking factor like this:

    static const Scalar inflate = (dir.array() == 0).any() ? 1 + 1e-10 : 1;

A function-local `static` is initialized exactly once, on the first call. The
`dir` of that very first call therefore fixes `inflate` for the rest of the
process, and every later box support query -- for every box, in every
direction -- silently reuses it.

The result is that queries which share no geometry, no transform and no
request still influence each other: what each one returns depends on which of
them ran first.

Since `inflate` is fixed once per process, a single process can only ever see
one of the two orders. This script therefore runs each order in its own
subprocess and puts the answers side by side.

Run it with coal's Python bindings importable:

    python support_function_statics.py
"""

import json
import subprocess
import sys

import numpy as np

import coal

# The two queries below are the same capsule-vs-box distance query, differing
# only in whether the box is rotated relative to the capsule.
#
# GJK starts from its default direction, (1, 0, 0). The capsule is the first
# shape and is asked for its support along that raw direction; the box is the
# second shape, and coal hands a second shape the direction rotated into its
# own frame. So the box sees a direction with no zero components when it is
# rotated (query A, which fixes `inflate` to 1) and the raw axis-aligned
# direction when it is not (query B, which fixes `inflate` to 1 + 1e-10).
ROTATED = np.array([0.7, 0.3, 0.5, 0.4])  # quaternion, (w, x, y, z)
UPRIGHT = np.array([1.0, 0.0, 0.0, 0.0])


def rotation_matrix(quaternion):
    """A rotation matrix from a (w, x, y, z) quaternion, which need not be
    normalized."""
    w, x, y, z = quaternion / np.linalg.norm(quaternion)
    return np.array(
        [
            [1 - 2 * (y * y + z * z), 2 * (x * y - w * z), 2 * (x * z + w * y)],
            [2 * (x * y + w * z), 1 - 2 * (x * x + z * z), 2 * (y * z - w * x)],
            [2 * (x * z - w * y), 2 * (y * z + w * x), 1 - 2 * (x * x + y * y)],
        ]
    )


def capsule_box_distance(quaternion):
    """The distance between a capsule and a box held at the given orientation.
    Every setting is left at its default."""
    capsule = coal.Capsule(0.3, 1.0)
    box = coal.Box(1, 1, 1)
    pose_capsule = coal.Transform3s()
    pose_box = coal.Transform3s(rotation_matrix(quaternion), np.array([3.0, 4.0, 5.0]))

    request = coal.DistanceRequest()
    result = coal.DistanceResult()
    return coal.distance(capsule, pose_capsule, box, pose_box, request, result)


def boxes_reported_in_contact(gap):
    """Does coal report two unit boxes, placed face to face and separated by
    `gap`, as being in contact? For any gap > 0 the true answer is "no"."""
    box1 = coal.Box(1, 1, 1)
    box2 = coal.Box(1, 1, 1)

    request = coal.CollisionRequest()
    request.num_max_contacts = 1
    request.security_margin = 0
    result = coal.CollisionResult()
    coal.collide(
        box1,
        coal.Transform3s(),
        box2,
        coal.Transform3s(np.eye(3), np.array([1.0 + gap, 0.0, 0.0])),
        request,
        result,
    )
    return result.isCollision()


# Where the reported contact boundary sits, given the box half side of 0.5:
#
#     gjk_tolerance + halfSide * (inflate - 1)
#
# That is `gjk_tolerance` when `inflate` is 1, and 5e-11 beyond it when
# `inflate` is 1 + 1e-10. Probing halfway between the two puts the boxes
# strictly outside each other either way, while landing on opposite sides of
# whichever boundary this process ended up with.
GJK_TOLERANCE = coal.CollisionRequest().gjk_tolerance
PROBE_GAP = GJK_TOLERANCE + 0.5 * 1e-10 / 2


def run_one_order(a_first):
    """Runs both queries in the given order, then classifies a pair of boxes
    that are definitely not touching. Meant to own its whole process."""
    if a_first:
        distance_a = capsule_box_distance(ROTATED)
        distance_b = capsule_box_distance(UPRIGHT)
    else:
        distance_b = capsule_box_distance(UPRIGHT)
        distance_a = capsule_box_distance(ROTATED)
    return {
        "distance_a": distance_a,
        "distance_b": distance_b,
        "in_contact": boxes_reported_in_contact(PROBE_GAP),
    }


def main():
    if len(sys.argv) > 1:
        # Child process: run one order and hand the numbers back.
        print(json.dumps(run_one_order(a_first=sys.argv[1] == "a_first")))
        return 0

    a_first, b_first = (
        json.loads(
            subprocess.run(
                [sys.executable, __file__, order],
                capture_output=True,
                check=True,
                text=True,
            ).stdout
        )
        for order in ("a_first", "b_first")
    )

    print(
        "coal picks the box support function's tie-breaking factor with a\n"
        "function-local `static`, so its value is fixed by the direction of\n"
        "the first box support query in the process, and reused by every\n"
        "query after it. See the module docstring for the details.\n"
    )
    print(f"{'':<44}{'A evaluated first':>22}{'B evaluated first':>22}")
    print(f"{'-' * 88}")
    for label, key in (
        ("query A: capsule vs rotated box", "distance_a"),
        ("query B: capsule vs upright box", "distance_b"),
    ):
        print(f"{label:<44}{a_first[key]:>22.17g}{b_first[key]:>22.17g}")
    label = f"boxes {PROBE_GAP:.6g} apart are"
    verdict = {True: "IN CONTACT", False: "separated"}
    print(
        f"{label:<44}{verdict[a_first['in_contact']]:>22}"
        f"{verdict[b_first['in_contact']]:>22}"
    )
    print()

    differs = [key for key in a_first if a_first[key] != b_first[key]]
    if not differs:
        print("Both orders agree. This build does not show the bug.")
        return 0

    print(
        "Each column is a separate process, running the same two independent\n"
        "queries in the opposite order. Every row should read the same down\n"
        f"the column. {len(differs)} of {len(a_first)} do not.\n"
    )
    print(
        "The last row is the plainest: the boxes are strictly separated, by\n"
        f"{PROBE_GAP:.6g}. Whether coal says so depends only on what ran before."
    )
    return 1


if __name__ == "__main__":
    sys.exit(main())

```

### Relevant log output / Error message

```shell
A evaluated first     B evaluated first
----------------------------------------------------------------------------------------
query A: capsule vs rotated box                 5.6165066132854538    5.6165066132052122
query B: capsule vs upright box                 5.5736700622353652    5.5736700621502395
boxes 1.00003e-06 apart are                              separated            IN CONTACT

Each column is a separate process, running the same two independent
queries in the opposite order. Every row should read the same down
the column. 3 of 3 do not.
```

### System Info

- Ubuntu 24.04
- pixi
- Built from eaa1b901
- Dependencies completely dictated by pixi

### Checklist

- [ ] I have checked that all my packages are installed from the same package manager (conda-forge, Nix, PyPI, robotpkg, ROS, ...)
- [x] I have checked that there is no similar [issue](https://github.com/coal-library/coal/issues)/[discussion](https://github.com/coal-library/coal/discussions) in the repo
- [x] I have provided a minimal and working example to reproduce the bug
- [x] I have used [Markdown code blocks](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) for code, stack traces and compiler errors
- [x] I have used an AI assistant

SeanCurtis-TRI:
## Description

Changes the `getShapeSupport(const Box*, ...)` method to remove the erroneous use of `static` in defining the `inflate` value. This corrects a small, subtle bug in which distance measurements (and even collision classification) can depend on the ordering of queries.

Incidentally, modifies the `getShapeSupport()` functions for `Cone` and `Cylinder` to use `constexpr` in place of `static const` to emphasize the compile-time `const`ness in contrast to `Box`'s run-time `constness`.

Fixes #894.

## Checklist

- [x] I have read the [contributing guidelines](../development/contributing.md)
- [x] I have run `pre-commit run --all-files` or `pixi run lint`
- [x] I have reviewed my own code
- [x] I have commented my code where necessary
- [X] I have made corresponding changes to the Doxygen documentation
- [X] I have added tests that prove my fix or feature works
- [x] I have updated the [CHANGELOG](https://github.com/coal-library/coal/blob/devel/CHANGELOG.md) or added the "no changelog" label for CI or infra changes
- [X] I have updated the [README credits section](https://github.com/coal-library/coal?tab=readme-ov-file#credits)
- [X] I have used an AI assistant


test/support_function_statics.cpp
SeanCurtis-TRI, line 1: To reviewers:

These tests represent my efforts to nail down the perceived bug. As such it includes both distance and collision errors. We can leave all of it intact or choose a smaller witness against regression. Simply doing the distance test would be sufficient (after all, if the distance values differ, that logically implies collision classification differences). I'd be happy to simplify the test to omit the recursive search for the collision boundary and keep just the distance tests.

README.md
```diff
@@ -285,6 +285,7 @@ In addition to the core dev team, the following people have also been involved i
 -   [Lars Berscheid](https://github.com/pantor): code cleanups
 -   Akseppal: bug fixes
 -   [Victor Engmark](https://github.com/l0b0): maintenance fixes
+-   [Sean Curtis]()https://github.com/SeanCurtis-TRI (Toyota Research Institute): code cleanups
```
lmontaut:
```suggestion
-   [Sean Curtis](https://github.com/SeanCurtis-TRI) (Toyota Research Institute): code cleanups
```

lmontaut:
Hi @SeanCurtis-TRI! Thanks a lot for spotting this mistake :)

For the test, something like this, using the support function directly, fails on the current `devel` but is fixed by your changes:
```cpp
  const Box box(2, 2, 2);
  int hint = 0;

  const Scalar inflated_half_side = Scalar(1) + Scalar(1e-10);
  const Vec3s directions[] = {Vec3s(-1, -1, -1), Vec3s(-1, 0, 0),
                              Vec3s(-1, -1, -1)};
  const Vec3s expected[] = {Vec3s(-1, -1, -1), Vec3s(-inflated_half_side, 0, 0),
                            Vec3s(-1, -1, -1)};

  for (int i = 0; i < 3; ++i) {
    const Vec3s support = getSupport(&box, directions[i], hint);
    for (int axis = 0; axis < 3; ++axis) {
      BOOST_CHECK_EQUAL(support[axis], expected[i][axis]);
    }
  }
```

README.md
SeanCurtis-TRI: Thanks

SeanCurtis-TRI:
@lmontaut Good call. Furthermore, it keeps me far away from the debug errors in CI. I'll be opening up an issue that subject separately.

test/support_functions.cpp
jorisv: Please, try to be less verbose. I don't know if you have the habit to write so much comment, but it really look like the typical LLM output. As stated in the [contribution guideline](https://github.com/coal-library/coal/blob/devel/development/contributing.md):

>  Never copy-paste AI-generated text into an issue, pull request description or comment.

test/support_functions.cpp
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+#define BOOST_TEST_MODULE COAL_SUPPORT_FUNCTION_STATICS
+#include <boost/test/included/unit_test.hpp>
+
+#include "coal/data_types.h"
+#include "coal/narrowphase/support_functions.h"
+#include "coal/shape/geometric_shapes.h"
+
+// The purpose of this test file is to accumulate miscellaneous tests against
```
jorisv: Coal is missing a support function unit test file. So let's say this file is to test `getShapeSupport` function.

test/support_functions.cpp
```diff
@@ -0,0 +1,94 @@
...
+// ways. By evaluating each in sequence, we confirm that each gets the correct
+// answer, regardless of what has been evaluated before it.
+//
+// If the inflation value ever gets "latched" in the support function, at least
+// one of the expected support values will not match the expectation.
+BOOST_AUTO_TEST_CASE(
+    box_support_function_inflate_regression_test,
+    *boost::unit_test::precondition([](boost::unit_test::test_unit_id) {
+      return kInflateIsObservable;
+    })) {
```
jorisv: I don't feel this precondition is useful. You can remove it.

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - Fix using NaN to initialize collision data (`Contact`, `CollisionResult`, `DistanceResult`). This prevents the absurd `Contact contact; contact == contact; // false` problem.
   - Fix NaNs coming from GJK/EPA when the algorithms (correctly) early stopped. NaNs indicate failure. In the case that GJK/EPA early stopped but ran fine, we set non-computed data to inf instead of NaN.
 - Fix `ContactPatchResult::swapObjects`: flip `ContactPatch::direction` along with the patch frame, and reflect every point of a patch instead of one of them, so `getNormal` reports the contact normal for shape-first queries against meshes and height fields ([#884](https://github.com/coal-library/coal/pull/884))
+- Fix the erroneous copypasta of using a `static const inflate` value for a conditional function-local variable in getShapeSupport(Box). ([#895](https://github.com/coal-library/coal/pull/895))
```
jorisv:
```suggestion
- Fix `Box` support function ([#895](https://github.com/coal-library/coal/pull/895))
```

jorisv (changes requested):

test/support_functions.cpp
SeanCurtis-TRI: "Less verbose" is a difficult value to hew to. I find that information bias infects code documentation. A terse statement that is clear to the coder at the time, may not be clear to anyone else (or even the programmer six months down the road). I value succinctness, but, by definition, while it implies concision, it also includes *completeness*.

Nevertheless, in deference to Coal's style, I'll attempt to rein it in and favor terse over complete.

test/support_functions.cpp
SeanCurtis-TRI: If this test were *ever* to be executed with `Scalar = float`, it will fail in a meaningless way.

I can appreciate the position that you guys have no intention of ever using `Scalar = float`, and, therefore, this precondition (and the supporting code/documentation) is more cost than benefit. That kind of trade off must always be considered.

I'd still propose leaving *some* acknowledgement of the fact that we know this test is not meaningful for any Scalar where 1.0000000001 is not a meaningful value. Perhaps simply a comment so that if anyone were ever to trip over it in the future, they won't have to reinvent the wheel in discovering why the test failed.

lmontaut: @SeanCurtis-TRI actually we do plan to use `float`, depending on the context. That's why we introduced `typedef double SolverScalar` which is used internally in gjk/epa, while the rest of coal can be `float` for example. 
The tests are written to pass with the `double` version of coal though, but ideally we should also test stuff using floats.

However, your remark remains true. In the support functions, we impose a `1e-10` value for the inflate thingy but it should actually be scalar-dependent...

SeanCurtis-TRI:
BTW I misspoke about the test *failing* for `Scalar = float`. It will pass without actually having done any work. Every equality test will have been satisfied because 1 will always equal 1 + 1e-10. So, the problem isn't an inscrutable failure, it's a meaningless success.

SeanCurtis-TRI:
Also, I'm not accustomed to using github's interface to do reviews. Github claims there's an unresolved "requested change" from @jorisv. I can't identify anything that is actionable for me. Please correct me if I've simply missed it.

test/support_functions.cpp
jorisv: Sorry @SeanCurtis-TRI I disagree with your level of verbosity. It doesn't help, at the reverse people will tend to not read this amount of text and will miss the important information.

As an example:

```
// The purpose of this test file is to accumulate miscellaneous tests against
// the getShapeSupport implementations. It doesn't aim to exhaustively prove
// correctness. Mostly, it will include tests that are added opportunistically
// based on recognition of previously missed errors.
```

Can be turned into

```
// Test getShapeSupport functions.
// This file will contains non regression test.
```

This also close this discussion: https://github.com/coal-library/coal/pull/895#discussion_r4081419119

jorisv:
@SeanCurtis-TRI Github claim there is unresolved request because I didn't approve the PR. It's a manual process so you must wait I'm available.

test/support_functions.cpp
SeanCurtis-TRI: I've taken your suggestion verbatim.

SeanCurtis-TRI:
> It's a manual process so you must wait I'm available.

I hope I didn't accidentally communicate impatience. That wasn't my intent. I could see that github was blocking on the fact that "jorisv Requested changes". My question was merely about user experience in the github tool. Other than github alluding to the fact that there was still a blocking issue, I couldn't figure out via the interface what that blocking issue *is*. My question was intended to find out if there's a way to use github review that I'm not aware to find out what that blocking request is.

Github's built-in review technology has historically been regarded as second rate and that's why I have never really used it.

That's actually one of the purposes of this otherwise trivial PR - to help me learn your code culture and tooling. So, I apologize if it seems like I came across impatient.

